Issue #30, opened 2015-01-05
Create a coherent diesel/dashboard flow for new users


mixonic:
* User creation and initial session open in diesel
* Redirect after creation to `/organizations/new` in dashboard
* After email verification, user is redirected into account/stack FTUX at diesel

mixonic:
Screenshots for the account/stack FTUX that must be ported to diesel:

1. [SSH Key](https://api.monosnap.com/image/download?id=NW7rNt6gIN8fcoKHScBI6dR6nrso1A)
2. [SSH Key error state](https://api.monosnap.com/image/download?id=f7EaSYfHEKVbkGqADVsk7guC39rp9s)
3. [App handle](https://api.monosnap.com/image/download?id=FEAQJsbEjXDk85BIgMU6ST3aGIi8Ul)
4. [Database handle](https://api.monosnap.com/image/download?id=GNir2x76rhgw9SwxeHx8AZW5dZvha2)
5. [Database handle and slider](https://api.monosnap.com/image/download?id=MlX9vuXrBSIdnu24hUDiZjxiRKVRIS)
6. [Provisioning poller](https://api.monosnap.com/image/download?id=0ygcvEQXQmqCDBeNbAlVFQDwSkqkxF)
7. [Success page](https://api.monosnap.com/image/download?id=jo19TShPvxxWYFefG89NMSdfYuzNGg)
8. User is directed to `apps.index`

sandersonet:
@mixonic, proposed changes from our discussion:

### 1. Pricing page

* Separate development and production boxes on pricing page
* Specific plans langing pages (future)

### 2. Create account

* CTA to get help or contact us
* Move organization name to account creation screen

### 3. Create organization

* Remove entirely (yay)

### 4. Create app and database

* Summary of what apps and databases are
* "Create" first app and databases

### 5. Activation email

* Brief description of what aptible is

### 6. Dashboard first-time

* CTA to deploy app created in Step 4
  * Prompts for user to enter SSH Key.  Brief instructions on how to get SSH key
  * Instructions on deploying via
* CTA to add people or roles

mixonic:
The new flow, page by page, would be:

1. Pricing page
2. Create account (with org name). Should have an alternative path available to just submit a query asking for more information.
3. Create App & Create DB on one page. This should also explain what an app and db are conceptually.
4. Submit a credit card. This is the only part of the flow that will be on the Ruby dashboard codebase.
5. Complete the email verification
[...]

PR #32, opened 2015-01-05
Redirect into index after user creation

mixonic:
Redirect the user into the application with a (crippled, read-only) session after user creation.

Incremental step toward #30. For you to review/merge @fancyremarker


fancyremarker:
This looks good to me. Thanks for the pairing! Not sure if you want to merge given the fact that the user won't be dropped immediately back into Diesel after user creation, but feel free to.

mixonic:
Merging this and continuing to iterate

Issue #28, opened 2015-01-05
Feature Stories


sandersonet:
### 1. Apps & databases

- [x] Operations history (infinite scroll?) (#29) (pagination of any kind is blocked by the API not supporting pagination)
- [x] Scale slider, add db type selector (in progress: #33. PR: #37)
- [x] Create disk when creating DB (#38 PR: #39 )

### 2. Roles & permissions

- [x] Verify Stack permissions when CRUD apps/dbs (#42)
  * Users with a 'Privileged' role, can read and manage all stacks inside Org
  * User with Read or Manage permission can see a Stack and its apps and dbs
  * User must have Manage permission to edit stack settings, create/edit apps and dbs

### 3. First-time UX

- [ ] First-time wizard (#30)
  *  Create SSH Key
  * Create App
  * Create Database
  * Configuration details

### 4. Deployment/CI
- [ ] Anything left here?

* * *

### 1. Auth

- [x] Account creation flow
- [ ] Organization creation
- [ ] Stack creation
- [ ] Password reset
- [ ] Invitation acceptance
- [ ] Organization switcher

Account on-boarding flow requires some Auth rework as we will have to move our Stripe integration from dashboard.aptible.com to auth.aptible.com.

### 2. User settings
- [ ] Profile
- [ ] Authentication settings: email/password
- [ ] SSH Key management

### 3. Organization settings
- [ ] Overview
  * Invitations CRUD
  * Member list
  * Billing Contact and Security Officer
  * Alert emails
- [ ] Stack management
- [ ] Role management
- [ ] Organization settings

sandersonet:
Updated to separate required features from optional features that can be handled by existing dashboard if necessary.  Features are roughly ordered by priority.

sandersonet:
Closing in favor of #51

PR #29, opened 2015-01-05
Show operations history for apps and databases

bantic:
refs #28 

 * Move database model from Pod to models/ dir, add hasMany 'operations'
 * Add databases.show.operations route, tests
 * Test for link to operations history on app and db index pages
 * Move audit trail to shared template

looks like:
diesel 2015-01-05 15-43-27


app/templates/shared/operations/-audit-history.hbs
```diff
@@ -0,0 +1,49 @@
+<div class="container">
+    <div class="row">
```
mixonic: Could use a nudge left

mixonic:
:+1: after that whitespace nudge. Excellent!

mixonic:
Force pushed a fix for the whitespace amended to the commit, mergin'.

PR #34, opened 2015-01-05
Pass sessions to dashboard

mixonic:

Using `withCredentials` allows the cookie returned when creating a user to be set at the top domain level. Then it can be used as a valid session at the dashboard server.

### DIFF
--- a/app/signup/route.js
+++ b/app/signup/route.js
@@ -1,7 +1,15 @@
 import Ember from 'ember';
+import config from "../config/environment";
 import { buildCredentials } from "../login/route";
+import { replaceLocation } from "../utils/location";
 
 export default Ember.Route.extend({
+
+  beforeModel: function(){
+    if (this.session.get('isAuthenticated')) {
+      this.replaceWith('index');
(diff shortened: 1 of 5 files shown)


mixonic:
if you start the dashboard server locally (with foreman) creating a new account will direct you to the organization/new page.

mixonic:
May need https://github.com/aptible/auth.aptible.com/pull/104 to be merged before it can be tested locally.

app/signup/route.js
```diff
@@ -20,8 +27,8 @@ export default Ember.Route.extend({
         // TODO: This option causes a user request to fetch data already present in the identity map
         return session.open('aptible', credentials);
       }).then(function(){
-        route.transitionTo('index');
-      });
+        replaceLocation([config.legacyDashboardHost, 'organizations/new'].join('/'));
+      }, function(){});
```
bantic: without a comment explaining that `session.open` failures are ok to be ignored (I assume that's the intent?), this empty function is a little confusing.

PR #39, opened 2015-01-07
Create a provision operation when creating a database

### LINKED ISSUE #38: create a disk with a new database
When the existing dashboard creates a new database, it [also creates a disk](https://github.com/aptible/dashboard.aptible.com/blob/master/app/controllers/databases_controller.rb#L23-L28) (by queueing a provision operation) if the db creation succeeds.

Need to update diesel to do the same thing.

cc @fancyremarker Does that sound right to you? Or should the API be modified to accept a `disk_size` along with a POST to create a new db, and have the API handle both in the same action?

refs #28

bantic:
* Add custom adapter for Operation that prepends /databases/:id when there is a database

fixes #38 refs #28

@mixonic for you


app/utils/build-url-with-prefix-map.js
```diff
@@ -0,0 +1,39 @@
+import Ember from 'ember';
+
+export default function buildURLWithPrefixMap(prefixPropertyMapping){
+  return function buildURL(type, id, record) {
+    var url = [],
+        host = Ember.get(this, 'host'),
+        prefix = this.urlPrefix();
```
mixonic: style-wise, multiple `var` statements are easier to debug (you can drop a breakpoint on any of them).

mixonic:
Gracias :-) :cake:

app/utils/build-url-with-prefix-map.js
bantic: yeah, good point. I lifted this directly from [ember-data's version](https://github.com/emberjs/data/blob/9e9a78111825d4154522cba8ddbbe679c56c9e5c/packages/ember-data/lib/adapters/rest_adapter.js#L532-L550), which looked like that. I'd prefer to minimize the diffs between this and that.

I'm gonna add in a link to the ember-data function, though.

mixonic: Sure. The no-mutiple-value-setting-var style is newish in Ember, but I've found it really useful. ember-data lags on best practices as usual :sob:

PR #40, opened 2015-01-07
Add support for email verification

mixonic:
@bantic for you to review/merge. /cc @fancyremarker we worked on this yesterday.

Adds support to diesel for email verification. This is still a stand-alone feature, and not part of the wizard flow.

Also adds a top level error page, required to show a default message for invalid tokens.

### DIFF
--- a/app/error/controller.js
+++ b/app/error/controller.js
@@ -1,0 +1,15 @@
+import Ember from "ember";
+
+export default Ember.Controller.extend({
+
+  title: function(){
+    return this.get('model.status') || this.get('model.name');
+  }.property('model.status', 'model.name'),
+
+  subTitle: Ember.computed.alias('model.statusText'),
+
+  message: function(){
+    return this.get('model.responseJSON.message')+'.' || 'Our JavaScript raised an exception.';
+  }.property('model.responseJSON.message')
+
+});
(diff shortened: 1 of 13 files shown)


app/error/controller.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default Ember.Controller.extend({
+
+  title: function(){
+    return this.get('model.status') || this.get('model.name');
+  }.property('model.status', 'model.name'),
+
+  subTitle: function(){
+    return this.get('model.statusText');
+  }.property('model.statusText', 'model.message'),
```
bantic: is it actually dependent on model.message?

app/models/verification.js
```diff
@@ -0,0 +1,6 @@
+import DS from 'ember-data';
+
+export default DS.Model.extend({
+  verification_code: DS.attr('string'),
```
bantic: Is there a reason for not having this camelCased `verificationCode`?

app/utils/url-parameters.js
```diff
@@ -0,0 +1,21 @@
+export function getUrlParameters(loc) {
+  var a = loc.search.substr(1).split('&');
+  var b = {};
+  if (a === "") {
```
bantic: wouldn't `a` be an array? And if there's no search it should be `[""]` then?

bantic:
@mixonic sweet — I added a few qs there, after you take a look at those go ahead and merge at your leisure

mixonic:
@bantic updated to address your comments, and added unit tests for the url params parser. Thanks!

PR #43, opened 2015-01-07
Add organization to account creation flow

mixonic:
Moves us toward https://github.com/aptible/diesel.aptible.com/issues/30#issuecomment-69032950

The organization name can now be specified when creating an account. This will let us bypass the organization creation screen for most users. There is a client-side validation that ensures the user cannot forget to specify an organization name. In a later PR we should expand that to check and see if an organization with that names exists.

### DIFF
--- a/tests/acceptance/authentication-test.js
+++ b/tests/acceptance/authentication-test.js
@@ -169,8 +169,12 @@ test('Creating an account directs to login', function() {
   var email = 'good@email.com';
   var password = 'correct';
   var userUrl = '/users/my-user';
+  var organization = 'Great Co.';
 
-  stubRequest('post', '/users', function(){
+  stubRequest('post', '/users', function(request){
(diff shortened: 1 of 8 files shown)


mixonic:
For @bantic to review and merge.

tests/acceptance/authentication-test.js
```diff
@@ -188,12 +192,37 @@ test('Creating an account directs to login', function() {
...
   andThen(function(){
     locationUpdatedTo('http://legacy-dashboard-host.com/organizations/new');
   });
 });
+
+test('Creating an account waits on a valid organization name', function() {
+  var email = 'good@email.com';
+  var password = 'correct';
+  var userUrl = '/users/my-user';
+  var organization = 'bad';
```
bantic: a better name for this var might be `tooShortOrgName`

bantic:
nice. cool to see ember-validations in action.

PR #46, opened 2015-01-08
Create an app/db creation page for the FTUX

mixonic:
Adds a page at `/welcome/first-app` that looks like:

This is also linked to as the next step after creating a new user. The values entered are stored in localStorage to be used for provisioning at a later step. Refs #30.

@bantic for you to review and merge!


bantic:
nice. looks good!

PR #47, opened 2015-01-09
Wwwizard

mixonic:
Introduce a basic payment page (largely unstyled).

Paired with https://github.com/aptible/auth.aptible.com/pull/105, this provides the ability to go from account creation to app/db naming to payment information and the creation of a subscription. For now, the user is directed to `index` after payment info.

### DIFF
--- a/app/utils/stripe.js
+++ b/app/utils/stripe.js
@@ -1,0 +1,39 @@
+/* globals Stripe */
+
+import jQuery from 'jquery';
+import Ember from 'ember';
+import config from '../config/environment';
+
+var loadPromise;
+
+function load(){
+  if(!loadPromise) {
+    if (window.Stripe) {
+      loadPromise = Ember.RSVP.resolve();
+    } else {
+      loadPromise = new Ember.RSVP.Promise(function(resolve, reject) {
+        jQuery.getScript('https://js.stripe.com/v2/').then(Ember.run.bind(null, resolve), Ember.run.bind(null, reject));
+      }).then(function() {
+        Stripe.setPublishableKey(config.stripePublishableKey);
+      });
+    }
+  }
+
+  return loadPromise;
+}
+
+function createStripeToken(options) {
+  return load().then(function() {
+    return new Ember.RSVP.Promise(function(resolve, reject) {
+      Stripe.card.createToken(options, function(status, result) {
+        if(result.error) {
+          Ember.run(null, reject, result.error);
+        } else {
+          Ember.run(null, resolve, result);
+        }
+      });
+    });
+  });
+}
+
+export { createStripeToken };
--- a/config/environment.js
+++ b/config/environment.js
@@ -12,7 +12,7 @@ module.exports = function(environment) {
     legacyDashboardHost: "http://localhost:3000",
 
     authTokenKey: '_aptible_authToken',
-
+    stripePublishableKey: 'pk_test_eiw5HXHTAgTwyNnV9I5ruCrA',
     replaceLocation: true,
 
     EmberENV: {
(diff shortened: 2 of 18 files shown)


config/environment.js
```diff
@@ -34,7 +34,8 @@ module.exports = function(environment) {
     contentSecurityPolicy: {
       'connect-src': "'self' http://localhost:4000 http://localhost:4001 ws://localhost:35729 ws://0.0.0.0:35729",
       'style-src': "'self' 'unsafe-inline'",
-      'img-src': "'self' http://www.gravatar.com https://secure.gravatar.com"
+      'img-src': "'self' http://www.gravatar.com https://secure.gravatar.com",
+      'script-src': "'self' https://js.stripe.com https://api.stripe.com"
```
bantic: :+1:

tests/acceptance/welcome/payment-info-test.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+});
+
+test('visiting /welcome/payment-info when not logged in', function() {
+  visit('/welcome/payment-info');
+
+  andThen(function() {
+    equal(currentPath(), 'login');
+  });
+});
```
bantic: after #48 is merged there will be a `expectRequiresAuthentication` helper that we can use here

tests/acceptance/welcome/payment-info-test.js
```diff
@@ -0,0 +1,84 @@
...
+  fillIn('[name=number]', cardNumber);
+  fillIn('[name=cvc]', cvc);
+  fillIn('[name=exp-month]', expMonth);
+  fillIn('[name=exp-year]', expYear);
+  fillIn('[name=zip]', addressZip);
+  click('button:contains(Save)');
+
+  andThen(function() {
+    equal(currentPath(), 'apps.index');
+  });
```
bantic: this test would probably benefit from an `expect(8)` at the beginning — it's not clear that `stripeMock.createStripeToken` will necessarily be called, and the test could still pass if it were never called. I.e., someone could change the internals of `app/utils/stripe.js` to avoid calling `stripeMock.createStripeToken`, but this test wouldn't fail.

app/utils/stripe.js
```diff
@@ -0,0 +1,45 @@
...
+import jQuery from 'jquery';
+import Ember from 'ember';
+import config from '../config/environment';
+
+var loadStripe;
+var stripeMock = {};
+
+if(config.env === 'test') {
+  loadStripe = Ember.RSVP.resolve;
+  createStripeToken = Ember.RSVP.resolve;
```
bantic: The use of test-only code here feels like a code smell but I'm having trouble figuring out a good way to avoid it. It would probably be better to at least have the exported stripeMock have the same API as the Stripe global, i.e.:

```javascript
var stripeMock = {
  setPublishableKey: Ember.K,
  card: {
    createToken: function(){ return Ember.RSVP.resolve(); }
  }
};
``` 

?

bantic: Maybe it would be better to export an object here, and use `loadStripe` internally:

```javascript
var loadPromise;
var loadStripe = function(){
  if (!loadPromise) { loadPromise = Ember.RSVP.Promise(/* jQuery.getScript, etc, returning the Stripe global */); }
  return loadPromise;
};
export default {
  createToken: function(options){
    return loadStripe().then(function(stripe){
       return stipe.card.createToken(options);
    });
  }
}
```

Then in tests you could `import Stripe from "../utils/stripe"` and stub out `Stripe.createToken` to be a resolving or rejecting promise, depending on what you are testing. ?

bantic:
@mixonic looks good — I had a few thoughts for minimizing the differences between stripe in test mode and normal mode. I think it should be possible to not have to check whether we are in test mode inside the stripe util code, which felt like a code smell to me. Back to you — lmk if you'd prefer it as is.

mixonic:
@bantic you were absolutely correct. While doing the analytics stuff I remembered the golden rules and cleaned this up.

When testing acceptance, you should always mock out the 3rd part API, not the internal API. You can do this by stomping on the global representing the API. Much better now.

Issue #42, opened 2015-01-07
Use roles and permissions to hide/show UI elements


bantic:
Refs #28.

Diesel needs to read the roles for a user and permissions for a stack and use those to determine whether to show certain UI elements like the "create app" and "create database" links.

For the moment it should be sufficient to determine on a per-stack basis whether a user has the "manage" permission or not. The "read" permission is implied by the fact that the API returned the stack at all, and the "manage" permission is the only current permission that is required to be able to create/edit stacks, apps and databases.

See https://github.com/aptible/api.aptible.com/issues/144 for more on permissions and how they will change in the future.

A (probably) incomplete list of places in the UI where elements are conditionally shown based on the manage permission:

  * Apps listing page  ![aptible apps 2015-01-07 11-54-27](https://cloud.githubusercontent.com/assets/2023/5649474/141a8e6c-9664-11e4-8213-604555907f37.jpg)
  * App show page  ![google 2b hangouts 2015-01-07 11-16-06](https://cloud.githubusercontent.com/assets/2023/5649492/588b018a-9664-11e4-809b-a784a507d61b.png)

  * DB Listing page  ![aptible databases 2015-01-07 11-56-02](https://cloud.githubusercontent.com/assets/2023/5649482/32f413d0-9664-11e4-9d63-bfa2cbb3fdd7.jpg)

bantic:
cc @sandersonet and/or @fancyremarker can you take a look at the above and lmk if I missed anything important?

sandersonet:
Looks great.  I believe your screenshots cover all possible scenarios.

On Wed, Jan 7, 2015 at 11:59 AM, Cory Forsyth <notifications@github.com>
wrote:

> cc @sandersonet <https://github.com/sandersonet> and/or @fancyremarker
> <https://github.com/fancyremarker> can you take a look at the above and
> lmk if I missed anything important?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aptible/diesel.aptible.com/issues/42#issuecomment-69052980>
> .
>



-- 
*Skylar Anderson*
https://www.aptible.com

fancyremarker:
Super. How hard is it now to determine a user's capabilities? We can probably afford to encapsulate some of this information in the token itself.

sandersonet:
That could save some HTTP requests.  Right now, the current process is like so:

* Fetch current users roles
* For each stack, fetch stack's permissions
[...]

PR #48, opened 2015-01-09
roles/permissions determine visibility of parts of UI

bantic:
__Do not merge__

 * create role and permission models
 * {{aptible-ability}} component
 * update version of ember-data-hal-adapter
 * change test helper `signIn` to push a stubbed current user into the store and set on session

Still to do

  * [x] fix failing tests at travis-ci
  * [x] get [changes to hal adapter merged](https://github.com/201-created/ember-data-hal-9000/pull/2)
  * [x] tests for User.can, {{aptible-ability}}
  * [x] add authenticated route that preloads permissions/roles
  * [x] determine if we can avoid loading roles from stack permissions since the API gives their URLs already
  * [x] PR changes to api to embed permissions in stacks (https://github.com/aptible/api.aptible.com/pull/147)

refs #42 #28

### DIFF
--- a/app/components/aptible-ability/component.js
+++ b/app/components/aptible-ability/component.js
@@ -3,7 +3,7 @@ import can from '../../utils/can';
 
 export default Ember.Component.extend({
   hasAbility: false, // cannot use `isVisible` with `tagName: ''`
-  tagName: '',
+  tagName: '', // tagName must be exactly '' to render this component with no surrounding tag
 
   user: null,
   scope: null,
@@ -15,9 +15,9 @@ export default Ember.Component.extend({
         stack = this.get('stack'),
         component = this;
 
-     Ember.assert("You must provide a user to aptible-ability", user);
-     Ember.assert("You must provide a scope to aptible-ability", scope);
-     Ember.assert("You must provide a stack to aptible-ability", stack);
+     Ember.assert("You must provide a user to aptible-ability", !!user);
+     Ember.assert("You must provide a scope to aptible-ability", !!scope);
+     Ember.assert("You must provide a stack to aptible-ability", !!stack);
 
      can(user, scope, stack).then(function(bool){
        if (component.isDestroyed) { return; }
--- a/app/components/aptible-ability/component.js
+++ b/app/components/aptible-ability/component.js
@@ -1,0 +1,28 @@
+import Ember from 'ember';
+import can from '../../utils/can';
+
+export default Ember.Component.extend({
+  hasAbility: false, // cannot use `isVisible` with `tagName: ''`
+  tagName: '',
+
+  user: null,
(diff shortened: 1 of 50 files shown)


app/components/aptible-ability/component.js
```diff
@@ -0,0 +1,28 @@
...
+
+  checkAbility: function(){
+    var user  = this.get('user'),
+        scope = this.get('scope'),
+        stack = this.get('stack'),
+        component = this;
+
+     Ember.assert("You must provide a user to aptible-ability", user);
+     Ember.assert("You must provide a scope to aptible-ability", scope);
+     Ember.assert("You must provide a stack to aptible-ability", stack);
```
mixonic: passing the boolean ala `!!user` would be preferred.

app/components/aptible-ability/component.js
```diff
@@ -0,0 +1,28 @@
...
+
+     can(user, scope, stack).then(function(bool){
+       if (component.get('isDestroyed')) { return; }
```
mixonic: Should not need a `get(` for `isDestroyed`. It is a property on the object and not observable.

app/components/aptible-ability/component.js
```diff
@@ -0,0 +1,28 @@
+import Ember from 'ember';
+import can from '../../utils/can';
+
+export default Ember.Component.extend({
+  hasAbility: false, // cannot use `isVisible` with `tagName: ''`
```
mixonic: This usage of a component feels awkward b/c you're basically writing an `if` helper, but I think your solution here is idiomatic. +1.

To do it differently (more efficiently or just less weird feeling) we would need to muck with the `render` method or write a raw HTMLBars helper.

app/stack/new-database/route.js
```diff
@@ -1,6 +1,8 @@
 import Ember from 'ember';
 
 export default Ember.Route.extend({
+  requireAuthentication: true,
```
mixonic: Can these be stripped since `true` is the default?

mixonic:
@bantic I'd like to see the `requireAuthentication: true` statements get dropped, but the rest is :+1: and please :shipit: when you are good.

app/stack/new-database/route.js
bantic: yes — I left them in as a reminder of sorts that `Ember.Route` has been reopened. I am on the fence about dropping them. If you think they are just extra verbosity I will drop them.

mixonic: If you prefer that authentication always be explicit, but don't want to use a base class, you could assert that the `requireAuthentication` property is not undefined. That seems ok.

As they stand, I would be confused about what the default behavior is from looking at usage.

PR #50, opened 2015-01-13
Add a query form to the sign up page, analytics service

mixonic:
Unstyled, but working:

TODO:

* [ ] Review the analytics events with @sandersonet. They must need some special property to appear in intercom.io as a message.


sandersonet:
@mixonic Is the vision here to leave the form in the actual UI, or is this just to test event tracking?

sandersonet:
Also, we didn't have the intercom integration enabled in Segment.io for dev.  It's enabled now so events should be logging correctly.

mixonic:
For new users (no email) looks like this:





Submitting the form or clicking the button opens chat. Feels good!

mixonic:
also fwiw Intercom sez use this for sending a message with an event: http://docs.intercom.io/Intercom-for-user-analysis/Tracking-User-Events-in-Intercom

But the flow I've implemented here feels good imo.

@sandersonet if you are +1, then I would just like a review by @bantic and this is good to go.

sandersonet:
We need to have the intercom widget always available here, so I think you need to `identify()` regardless of email.  Other than that I think this seems like a good feature.

I will have to spend more time thinking about how to design this page further.  At first glance, having two email fields in such proximity seems confusing.  It may be best to bury the feature either behind a button click or blur/exit attempt as we mentioned earlier today.

Also, we need to be track a `page()` for each route in the app.  I don't explicitly have a problem adding it manually to every route, but it may be nice to have something more automatic.

bantic:
LGTM. nice. As @sandersonet mentions it would be nice to add an `.on('didTransition')` handler in the application route that tracks every "page"-view rather than doing it explicitly.

mixonic:
@sandersonet in the `app/application/route.js` file we call `identify`, which means it will be called for every time the app loads (and only once, since the application route can only be entered once).

I'd like to defer the page tracking to another PR, but it can definitely be automated :-)

@sandersonet I'd like to merge this and make a separate PR for styling (the cc page needs love as well). Unless you think the functionality itself is in question. Anyway, if it isn't ready for merging as-is, we should identify how to get to a better place. We can chat over it tomorrow.

sandersonet:
:+1:

Issue #51, opened 2015-01-13
Current Priorities


sandersonet:
### Completed

- [x] Invite acceptance functionality (@mixonic) https://github.com/aptible/diesel.aptible.com/pull/125
- [x] Signup flow
  - [x] Should restore users to previous state.  E.g. if user exists before submitting payment information, they should not be able to skip to stacks/show
  - [x] Currently, a stack is never actually created for a user when they enter payment information.  After a subscription is successfully created, we should also create a stack that matches the subscription plan.  We may need to update API to validate that stacks created have a valid type that matches the subscription level of the user.  E.g. If a user has only a development subscription, they shouldn't be able to provision a production stack.
    - [x] there is not form field for creating a stack. use the org name? Use the app name? - Currently we create a stack with the name `${organization.name} ${stack.type}`
  - [x] database should be provisioned after verification. App and db should be created after credit card data is entered (even before maybe, whatever is easier)

- [x] Empty app view (screen 203) (#78)
    -  [x] An un-deployed app (app with no services) should show quick checklist for deploying their app.  
    -  [x] Link out to manage SSH keys if user has none configured

- [x] [Improved styling#52](https://github.com/aptible/diesel.aptible.com/issues/52)
    - [x] Sign up flow styling
    - [x] New header and sub header styling
    - [x] App/DB Index
    - [x] App/DB Create
    - [x] Service/VHost index
    - [x] Service/VHost create
    - [x] App/DB Activity
    - [x] App/DB Deprovision

- [x] Refactor existing ember app to match new design's architecture (screens 205 - 214)
    - [x] Apps/DB index and new are now child outlets to Stack:show
    - [x] Services, Vhosts, Activity, and Deprovision are child outlets to App:show
    - [x] Activity and Deprovision are child outlets to Database:show
    - [x] Soon stacks will return rate information from API.  Consider how we'll use this with current container configuration to calculate an estimated cost of each service. (screen 212)
    - [x] service list page (without rate information, for now) (in progress: https://github.com/aptible/diesel.aptible.com/issues/62)

- [x] Paginated resources (screen 216)
    -  [x] Currently only applies to App/DB activity, but has bearing on eventual pagination of other resources and possibly compliance. (PR: #60)

- [x] User settings (design unchanged)
    - [x] SSH Keys display page (#66)
    - [x] Email/Password display page (#66)
    - [x] Name/Gravatar display page (#66)
    - [x] SSH Key page allows user to add/remove keys
[...]

PR #53, opened 2015-01-13
Pagination for apps and db operations pages

### LINKED ISSUE #49: display pagination for app/ and db/ operations lists
The API will return pagination info for collection resources (see https://github.com/aptible/api.aptible.com/pull/146). Diesel should use these to display pagination links, starting with the `apps/:id/operations and` and `databases/:id/operations` history pages.

bantic:
@mixonic for your review

  * Override findQuery for operation adapter to allow pagination
  * add pagination in {{aptible-pagination}} component
  * Paginated route and controller mixins for shared behaviors
  * update 'page' query param in url
  * dry tests with shared test helper functions

fixes #49 refs #51

cc @fancyremarker

This uses the current_page, total_count and per_page meta data
provided by the API as of https://github.com/aptible/api.aptible.com/pull/146

### DIFF
--- a/app/apps/show/operations/controller.js
+++ b/app/apps/show/operations/controller.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
+import Paginated from 'diesel/mixins/controllers/paginated';
 
-export default Ember.Controller.extend({
+export default Ember.Controller.extend(Paginated, {
   operationResourceType: 'app'
 });
--- a/app/databases/show/operations/route.js
+++ b/app/databases/show/operations/route.js
@@ -1,14 +1,19 @@
 import Ember from 'ember';
+import Paginated from "diesel/mixins/routes/paginated";
 
-export default Ember.Route.extend({
-  model: function(){
+export default Ember.Route.extend(Paginated, {
+  model: function(params, transition){
     var db = this.modelFor('databases.show');
+    var page = parseInt(transition.queryParams.page || 1, 10);
 
-    return db.get('operations');
+    return this.store.find('operation', {database:db, page:page});
   },
 
   setupController: function(controller, model){
-    controller.set('db', this.modelFor('databases.show'));
+    var db = this.modelFor('databases.show');
     controller.set('operations', model);
+    controller.set('paginatedResourceOwner', db);
+
+    this.setPaginationMeta(controller, 'operation');
   }
 });
--- a/app/databases/show/operations/route.js
+++ b/app/databases/show/operations/route.js
@@ -2,9 +2,9 @@ import Ember from 'ember';
 import Paginated from "diesel/mixins/routes/paginated";
 
 export default Ember.Route.extend(Paginated, {
-  model: function(params, transition){
+  model: function(params){
     var db = this.modelFor('databases.show');
-    var page = parseInt(transition.queryParams.page || 1, 10);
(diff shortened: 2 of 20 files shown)


app/apps/show/operations/controller.js
```diff
@@ -1,5 +1,6 @@
 import Ember from 'ember';
+import Paginated from 'diesel/mixins/controllers/paginated';
```
mixonic: relative imports are generally preferred, as they avoid sprinkling the app name around and make the code more portable. At least, this has been the idiom so far.

app/databases/show/operations/route.js
```diff
@@ -1,14 +1,19 @@
 import Ember from 'ember';
+import Paginated from "diesel/mixins/routes/paginated";
 
-export default Ember.Route.extend({
-  model: function(){
+export default Ember.Route.extend(Paginated, {
+  model: function(params, transition){
     var db = this.modelFor('databases.show');
+    var page = parseInt(transition.queryParams.page || 1, 10);
```
mixonic: I understood that query params are always coerced to the same type as their original value (in this case `1` in the paginated mixin).

Also I believe the query param is on `params`, and accessing the transition is unnecessary. I can confirm.

app/mixins/routes/paginated.js
```diff
@@ -0,0 +1,39 @@
...
+    goToPage: function(page){
+      var controller = this.controller;
+      var store = this.store;
+      var route = this;
+      var query = { page: page };
+
+      var paginatedResourceOwner = this.controller.get('paginatedResourceOwner');
+      query[ paginatedResourceOwner.constructor.typeKey ] = paginatedResourceOwner;
+
+      store.find('operation', query).then(function(operations){
```
mixonic: If this is a generic mixin, it should likely not reference `'operation'`

app/apps/show/operations/controller.js
bantic: Agreed, although in some cases I find the absolute paths like these easier to use intra codebase. The relative paths are non-copy-pastable if the files are in different folder paths. I.e., to use Paginated in the apps/index route you'd `import Paginated from "../../mixins/routes/paginated"` but for the apps/operations/index it is `import Paginated from "../../../mixins/routes/paginated"`, while `import Paginated from "diesel/mixins/routes/paginated"` works for both. The volatility of import namespace (diesel) feels lower to me than volatility of folder nesting. But I am not religious about this.

app/mixins/routes/paginated.js
mixonic: I might expect that the query to load a new page is a hook called on the route. So a route must include the mixin, and define a hook for fetching new pages. That way all the `store` related work stays on the route and this mixin is only responsible for managing convenience actions (next/prev page) and passing params to hooks on the route.

app/apps/show/operations/controller.js
mixonic: Aye. I would like to ponder a heuristic for when to use absolute vs relative.

app/databases/show/operations/route.js
bantic: that's correct — the queryParam is coerced into params as its property name. In this case `params.currentPage`. I'll push a commit in a second.

bantic:
closed in favor of #60

PR #58, opened 2015-01-15
Update styles and flow on create apps page

mixonic:
@bantic for you to review/merge. Pretty simple.

Adds styles and cancel functionality to the app creation page.

Also filters the count of apps and databases by if they are not dirty. This keeps the count from going up by one when `createRecord` is called. It may break in the case of edit (making it go down by one) but let us cross that when we get to it. Probably just a better prop to filter on.


app/stack/controller.js
```diff
@@ -1,5 +1,9 @@
 import Ember from "ember";
 
 export default Ember.Controller.extend({
-  hasMultipleStacks: Ember.computed.gt('stacks.length', 1)
+  hasMultipleStacks: Ember.computed.gt('stacks.length', 1),
+
+  persistedApps: Ember.computed.filterBy('model.apps', 'isDirty', false),
+  persistedDatabases: Ember.computed.filterBy('model.databases', 'isSaved', false)
```
bantic: What's the reason to use `isSaved` for dbs and `isDirty` for apps?

bantic:
@mixonic looks good. I'm curious about the `isSaved` vs `isDirty` for persisted apps/dbs CPs, but otherwise good to merge imo

mixonic:
@bantic `isSaved` is not a flag on models. I looked at some source and it seems `isNew` is what I really want to filter by. Updated here, and I've also added the component for matching sub-routes.

PR #63, opened 2015-01-16
Add slider for services of an app

mixonic:

Adds the sliders to the containers of an application, with some scratched-up UI (not final).

### DIFF
--- a/app/app/services/route.js
+++ b/app/app/services/route.js
@@ -5,6 +5,17 @@ export default Ember.Route.extend({
   model: function(){
     var app = this.modelFor('app');
     return app.get('services');
+  },
+
+  actions: {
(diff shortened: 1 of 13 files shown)


bantic:
@mixonic looks good. I'm looking into the source of the test error at travis

bantic:
@mixonic Added a commit to bring slider style more inline, and made a UX change to only show the scale/cancel buttons after the user finishes dragging the slider. Looks like:

mygsyrsux2

The build continues to error at travis, but I pushed an identical branch to my remote. If it passes I will merge. You may need to clear your caches at travis?

bantic:
This [passes at Travis for me](https://travis-ci.org/bantic/diesel.aptible.com/builds/47267678): http://take.ms/TdCW5

Merging.

Issue #67, opened 2015-01-16
user dropdown menu is underneath nav bar


bantic:
The dropdown menu is obscured by the nav bar:

![diesel 2015-01-16 18-53-37](https://cloud.githubusercontent.com/assets/2023/5786327/1b49b344-9db1-11e4-8354-80981308b37b.jpg)

bantic:
closed by updating aptible-sass in https://github.com/aptible/aptible-sass/pull/64

Issue #66, opened 2015-01-16
implement user settings & ssh key pages (design unchanged) in diesel


bantic:
Implement the existing dashboard settings routes `settings/ssh` and `settings/admin`

refs #51 

They look like this in dashboard now:

![aptible cory forsyth account settings 2015-01-16 12-45-01](https://cloud.githubusercontent.com/assets/2023/5781073/b22cb2a4-9d7d-11e4-9053-672218146835.jpg)

![aptible cory forsyth ssh keys 2015-01-16 12-45-12](https://cloud.githubusercontent.com/assets/2023/5781078/b7fae142-9d7d-11e4-8dc8-f63a026ec71c.jpg)

PR #68, opened 2015-01-17
Add settings profile, admin and ssh key pages

bantic:
The ajax is not yet wired up for editing name, password,
email or adding ssh keys.

refs #51 #66

The pages are now working for displaying data. The link to settings is obscured by the navbar (#67) but you can go directly to `/settings` to see the pages.

Notes for wiring up the data are in a separate issue.

This PR makes the pages look like:

fdiu2tp6fj

@sandersonet for you cc @mixonic

### DIFF
--- a/app/adapters/ssh-key.js
+++ b/app/adapters/ssh-key.js
@@ -1,0 +1,31 @@
+import AuthAdapter from './auth';
+import buildURLWithPrefixMap from '../utils/build-url-with-prefix-map';
+import Ember from 'ember';
+
+export default AuthAdapter.extend({
+  buildURL: buildURLWithPrefixMap({
+    'users': 'user.id'
+  }),
+
+  // To delete an ssh-key, the adapter should send a DELETE to /ssh_keys/:id,
+  // but it POSTs to /users/:user_id/ssh_keys.
+  // Override deleteRecord to remove the "/users/:user_id" from the
+  // url for a DELETE.
+  deleteRecord: function(store, type, record){
+    var id = Ember.get(record, 'id');
+    var userId = Ember.get(record, 'user.id');
+
+    var url = this.buildURL(type.typeKey, id, record);
+
+    if (userId) {
+      url = url.replace("/users/" + userId, "");
+    }
+
+    return this.ajax(url, "DELETE");
+  },
+
+  // In URLs and JSON payloads, use "ssh_keys" instead of "sshKeys"
+  pathForType: function(type){
+    return Ember.String.pluralize( Ember.String.decamelize(type) );
+  }
+});
--- a/app/settings/admin/template.hbs
+++ b/app/settings/admin/template.hbs
@@ -1,0 +1,84 @@
+<div class="panel-group">
+  <div class="panel panel-default">
+    <div class="panel-heading"><h3>Change Your Password</h3></div>
+    <div class="panel-body">
+      {{#if passwordError}}
+        <div class="alert alert-warning">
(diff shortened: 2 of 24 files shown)


bantic:
Still to do:

  * [ ] DRY ssh tests
  * [ ] allow user to delete ssh key
  * [ ] better styling for ssh adding, form, errors
  * [ ] test that create SSH key posts to /users/:user_id/ssh_keys

bantic:
Still to do:

 * [ ] change password
 * [ ] change email
 * [ ] change name

bantic:
This is ready to go now.
The user can add/delete SSH keys and change name, email and password.
@mixonic ready for your review.
This depends on changes to auth that are in https://github.com/aptible/auth.aptible.com/pull/109.

Some screencasts:

  * change name:
68xhpkqxhm

  * change email:
email

  * add/delete SSH key
sshkey

bantic:
The dependent PR in auth has been merged, so this PR is no longer blocked on anything.

app/adapters/ssh-key.js
```diff
@@ -0,0 +1,31 @@
...
+    var userId = Ember.get(record, 'user.id');
+
+    var url = this.buildURL(type.typeKey, id, record);
+
+    if (userId) {
+      url = url.replace("/users/" + userId, "");
+    }
+
+    return this.ajax(url, "DELETE");
+  },
```
mixonic: Sweet. I will take this and maybe push it up into application or auth adapter if I can get the logic right.

app/settings/admin/template.hbs
```diff
@@ -0,0 +1,84 @@
...
+        <div class="alert alert-warning">
+          {{passwordError}}
+        </div>
+      {{/if}}
+      <form action="">
+        <div class="form-group">
+          <label for="password">New password</label>
+          {{input type="password"
+                  class="form-control" placholder="New password"
+                  value=session.currentUser.password name="password"}}
```
mixonic: I might suggest returning `session.currentUser` as the model of this route, then having the form and actions refer to that. Seems more easily refactored.

tests/unit/models/ssh-key-test.js
```diff
@@ -0,0 +1,88 @@
...
+    'model:token',
+
+    'adapter:application',
+    'adapter:ssh-key',
+    'adapter:user',
+    'serializer:application',
+  ]
+});
+
+/*
```
mixonic: Can you add a reason to leave these tests commented?

app/adapters/ssh-key.js
bantic: I was thinking that maybe `buildURLWithPrefix` could take a map of verbs -> prefixes, so it would be, like: `buildURLWithPrefix({'POST': {'users': 'user.id'}});`. Not sure how difficult that would be but it probably makes some sense, since the paths are definitely verb-dependent. I am pretty sure the GET for ssh_keys is also non-prefixed (like the DELETE).

tests/unit/models/ssh-key-test.js
bantic: oops, this was mistakenly committed. I will remove it, instead.

app/settings/admin/template.hbs
bantic: I'll give that a shot.

bantic:
@mixonic all set now

mixonic:
I've rebased this and force pushed it to your branch @bantic. Will merge when the tests go green

PR #70, opened 2015-01-26
Services and VHost styles

sandersonet:
* Updated to new micro footer
* Update slider styles
* Adding cost estimator component to service sliders
* Updated vhost styles

Requires an updated aptible-sass#diesel

screen shot 2015-01-26 at 12 38 55 pm
screen shot 2015-01-26 at 12 39 08 pm


sandersonet:
For you @mixonic

app/components/estimated-cost/component.js
```diff
@@ -0,0 +1,26 @@
...
+
+  total: function() {
+    return this.centsToDollars(
+      this.get('hoursPerMonth') * this.get('count') * this.get('stack.appContainerCentsPerHour')
+    );
+  }.property('stack.appContainerCentsPerHour', 'count'),
+
+  centsToDollars: function(cents) {
+    return '$' + (cents/100).toFixed(2);
+  }
```
mixonic: these functions can just be private methods in the file before the class definition. ES6 modules give us a private namespace for this file :-D

app/components/estimated-cost/component.js
```diff
@@ -0,0 +1,26 @@
...
+
+    var p = this.get('count') === 1 ? '' : 's';
+    return this.get('stack.type').capitalize() + ' App Container' + p;
+  }.property('stack.type', 'count'),
+
+  total: function() {
+    return this.centsToDollars(
+      this.get('hoursPerMonth') * this.get('count') * this.get('stack.appContainerCentsPerHour')
+    );
+  }.property('stack.appContainerCentsPerHour', 'count'),
```
mixonic: This should have `hoursPerMonth` as a dependent key

mixonic: Ah I see, it is a static value. In that case I would suggest just making it a local value defined before the class, or accessing it as `this.hoursPerMonth`

app/components/service-scaler/component.js
```diff
@@ -18,6 +18,11 @@ export default Ember.Component.extend({
     this.set( 'containerCount', this.get('service.containerCount') );
   }.on('init').observes('service.containerCount'),
 
+  unitOfMeasure: function() {
+    var type = this.get('service').get('stack').get('type');
```
mixonic: `this.get('service.stack.type')`

app/serializers/application.js
```diff
@@ -19,6 +19,10 @@ export default HalSerializer.extend({
   normalize: function(type, hash, property) {
     var payload = this._super(type, hash, property);
 
+    if(payload["default"]) {
+      payload["isDefault"] = payload["default"];
```
mixonic: Oof, this seems quirky. To make a single point, only data coming from the server will be changed to `isDefault`. When a record is saved, `isDefault` will not be mapped back to `default`.

I suggest using `attrs` on a serializer:

http://emberjs.com/api/data/classes/DS.RESTSerializer.html#property_attrs

This would allow you to be very specific here. In `app/serializers/vhost.js`:

```js
// ...
  attrs: {
    isDefault: 'default'
  }
```

This way only vhost has the mapping.

mixonic:
@sandersonet great tests! I've left some small additional feedback.

sandersonet:
Thanks for the feedback @mixonic, PR updated with recommended changes.

mixonic:
Great stuff @sandersonet

PR #75, opened 2015-01-27
Signup flow styles

sandersonet:
Improved styles for
* Login
* Signup
* Create app
* Add payment info

Latest on aptible-sass#diesel required.

### DIFF
--- a/app/welcome/first-app/route.js
+++ b/app/welcome/first-app/route.js
@@ -1,23 +1,8 @@
 import Ember from 'ember';
-import { read, write } from '../../utils/storage';
+import { write } from '../utils/storage';
 export var firstAppKey = '_aptible_firstAppData';
 
 export default Ember.Route.extend({
-  model: function(){
-    var firstApp = read(firstAppKey);
-
-    if (firstApp) {
-      return firstApp;
-    } else {
-      return {
-        appHandle: '',
-        dbHandle: '',
-        diskSize: 10,
-        dbType: null
-      };
-    }
-  },
-
   setupController: function(controller, model) {
     controller.set('model', model);
   },
--- a/app/welcome/first-app/route.js
+++ b/app/welcome/first-app/route.js
@@ -1,8 +1,23 @@
 import Ember from 'ember';
-import { write } from '../../utils/storage';
+import { read, write } from '../../utils/storage';
 export var firstAppKey = '_aptible_firstAppData';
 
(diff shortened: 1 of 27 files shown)


sandersonet:
For you @mixonic

app/welcome/first-app/route.js
```diff
@@ -1,23 +1,8 @@
...
-      return firstApp;
-    } else {
-      return {
-        appHandle: '',
-        dbHandle: '',
-        diskSize: 10,
-        dbType: null
-      };
-    }
-  },
```
mixonic: This code was responsible for restoring the app and db as entered if a user returned to the wizard. And in general it provides defaults for the initial screens.

It seems like it should not be dropped?

sandersonet: I moved it to the parent route.

app/welcome/payment-info/route.js
```diff
@@ -2,9 +2,6 @@ import Ember from 'ember';
 import { createStripeToken } from '../../utils/stripe';
 
 export default Ember.Route.extend({
-  model: function() {
-    return {};
-  },
```
mixonic: This was also intentional- to return a blank object.

app/welcome/payment-info/controller.js
```diff
@@ -0,0 +1,7 @@
+import Ember from "ember";
+
+export default Ember.Controller.extend({
+  months: ['01', '02', '03', '04', '05', '06', '07', '08', '09', '10', '11', '12'],
+  years: ['2015', '2016', '2017', '2018', '2019', '2020'],
+  plans: ['development', 'pilot', 'platform', 'production']
```
mixonic: When details like this are stored on a controller, I usually end up creating components. For example, `months` is used in a select box. So I suggest a `select-months` component that internalizes the list of options. This is much more re-usable and semantic.

app/welcome/first-app/route.js
mixonic: That means the app/db info object is now shared with the payment info object? What is the motivation? IMO it is more flexible/isolated to keep them separate.

sandersonet:
@mixonic Updated per your feedback.  I'm still not sure about the `model` function in `welcome/payment-info/route.js`.  The goal here is to be able to use the app/db name created in the prior step in the copy on the right.  Thoughts?  Thanks for your help.

PR #76, opened 2015-01-28
Deprovision pane for apps

mixonic:
You can deprovision (destroy) an app. @fancyremarker this just destroys the app, let me know if this should be doing anything more fancy.

Success takes you back to the stack view.

### DIFF
--- a/app/adapters/application.js
+++ b/app/adapters/application.js
@@ -35,6 +35,16 @@ export default HalAdapter.extend({
     } else {
       return error;
     }
+  },
+
+  deleteRecord: function(store, type, record){
+    var id = Ember.get(record, 'id');
+    var url = this.buildURL(type.typeKey, id);
+    return this.ajax(url, "DELETE");
+  },
+
+  find: function(store, type, id /*, record */) {
+    return this.ajax(this.buildURL(type.typeKey, id), 'GET');
   }
 
 });
--- a/app/app/deprovision/route.js
+++ b/app/app/deprovision/route.js
@@ -1,0 +1,29 @@
+import Ember from "ember";
+
+export default Ember.Route.extend({
+
+  actions: {
+    deprovision: function(){
+      var app = this.currentModel;
+      var route = this;
+      var controller = this.controller;
+      var store = this.store;
+      controller.set('error', null);
+
+      app.get('stack').then(function(stack) {
+        var op = store.createRecord('operation', {
+          type: 'deprovision',
+          app: app
+        });
+        return op.save().then(function(){
+          return app.reload();
+        }).then(function(){
+          route.transitionTo('stack', stack);
+        }, function(e){
+          controller.set('error', e);
+        });
+      });
+    }
+  }
+
+});
--- a/app/app/deprovision/template.hbs
+++ b/app/app/deprovision/template.hbs
@@ -1,1 +1,35 @@
-Deprovision {{model.handle}}
+<div class="row">
+  <div class="col-xs-12">
+    <div class="panel panel-default panel-subdued">
+      <div class="panel-heading">
+        <h3>Permanently deprovision {{model.handle}} </h3>
+      </div>
+      <div class="panel-body">
+        <form role="form">
+          {{#if error}}
+            <div class="alert alert-warning fade in">
+              <a class="close" data-dismiss="alert">×</a>
+              <p>
+                There was a problem deprovisioning your application: {{error.responseJSON.message}}
+              </p>
+            </div>
+          {{/if}}
+
+          <p>
+            This action CANNOT be undone. This will permanently deprovision
+            <strong>{{model.handle}}</strong>
+          </p>
+          <p>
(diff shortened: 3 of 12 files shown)


app/adapters/application.js
```diff
@@ -35,6 +35,12 @@ export default HalAdapter.extend({
     } else {
       return error;
     }
+  },
+
+  deleteRecord: function(store, type, record){
```
mixonic: @bantic I think this should be all that is needed to force deletes to hit the non-nested route. I just don't pass the record into `buildURL`, so there is no nesting.

fancyremarker:
Ah — this should actually be creating an operation on the app with `{ type: "deprovision" }` (no other arguments necessary.

mixonic:
@fancyremarker should it also destroy? We can chat over it in detail this afternoon and fix the logic before this merges.

fancyremarker:
Negative. Once the operation is executed by Sweetness, Sweetness will take care of destroying the app.

The issue here is that there are a bunch of resources that need to be cleaned up for an app (app containers, proxy containers, logging containers, Elastic Load Balancer(s), etc.) and so Sweetness needs to do that before the app can actually be deleted. Moreover, we currently have different policies for exactly when the `deprovision` operation gets handled by Sweetness: development apps are deprovisioned immediately, while production apps sit on the queue until manually reviewed.

In the final version of our API, a non-Sweetness user won't even be able to call `DELETE /apps/:id` but that's [a separate issue](https://github.com/aptible/api.aptible.com/issues/103).

app/adapters/application.js
mixonic: I don't actually use this deletion code anymore. Instead this flow goes through deprovision

mixonic:
Ready for review by @bantic. Works best with https://github.com/aptible/api.aptible.com/pull/160

app/app/deprovision/route.js
```diff
@@ -0,0 +1,29 @@
...
+      var store = this.store;
+      controller.set('error', null);
+
+      app.get('stack').then(function(stack) {
+        var op = store.createRecord('operation', {
+          type: 'deprovision',
+          app: app
+        });
+        return op.save().then(function(){
+          return app.reload();
```
bantic: this is to pick up the fact that the status of the app is now 'deprovisioning', yeah? Is that info shown in the UI anywhere?
Not strictly necessary, but the intent of the [app model STATUSES](https://github.com/aptible/diesel.aptible.com/blob/master/app/models/app.js#L4-L7) variable is to enumerate all the possible valid statuses. Might be worth adding "deprovisioning" there.

app/app/deprovision/template.hbs
```diff
@@ -1 +1,35 @@
...
+
+          <p>
+            This action CANNOT be undone. This will permanently deprovision
+            <strong>{{model.handle}}</strong>
+          </p>
+          <p>
+            <strong>Type the name of the app to permanently deprovision.</strong>
+          </p>
+          <div class="form-group">
+            {{#button-with-confirmation confirmValue=model.handle action='deprovision'}}
```
bantic: this should probably have a class of "btn-danger" to make it red (as shown in sketch: http://take.ms/umRpm)

app/components/button-with-confirmation/component.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  isConfirmed: null,
+  isUnconfirmed: Ember.computed.not('isConfirmed'),
+  input: function(){
+    var enteredValue = this.get('enteredValue');
+    var confirmValue = this.get('confirmValue');
+    this.set('isConfirmed', enteredValue === confirmValue);
+  }.on('init'),
```
bantic: Might be clearer to make `isConfirmed` a cp?

```
isConfirmed: function(){
  return this.get('enteredValue') === this.get('confirmValue');
}.property('enteredValue', 'confirmValue')
```

I had actually forgotten that `input` was an event handler, so I didn't realize what was happening here.

tests/acceptance/apps-deprovision-test.js
```diff
@@ -0,0 +1,108 @@
...
+      type: 'deprovision'
+    });
+  });
+
+  stubStacks();
+
+  signInAndVisit('/apps/'+appId+'/deprovision');
+  fillIn('input[type=text]', appName);
+  andThen(function(){
+    $('input[type=text]').trigger('input');
```
bantic: If the `button-with-confirmation`'s `isConfirmed` is a CP, this manual trigger appears to not be necessary.

bantic:
awesome — LGTM. A few notes, merge at will.

mixonic:
@bantic This component is definitely a little different than what we've sometimes done in the past. I wrote it experimenting with the idea of data-down actions-up, thus my `input` driven logic instead of a CP dependent on the input value. I'd like to keep experimenting to see if we can make this pattern scale up.

PR #77, opened 2015-01-28
Updates HAL adapter to 0.1.5

bantic:
The 0.1.5 version of the HAL adapter now uses the serializer's `typeForRoot`
and the adapter's `pathForType`. It also removes the `serializeIntoHash`
override, because that's not part of HAL. This PR puts that back in
to diesel's application serializer.

Also removes `attrs` in the application serializer -- I do not believe these
are used.

@mixonic for you.


bantic:
please wait until #68 is merged

bantic:
will reopen after #68

PR #81, opened 2015-01-30
hal adapter -> 0.1.6

bantic:
* v0.1.5 removed the `serializeIntoHash` method, this PR adds it to the application serializer
  * v0.1.6 will load embedded resources with custom names, like
  * `lastOperation` and `lastDeployOperation` on app


mixonic:
Woo!

PR #82, opened 2015-01-30
Reload apps and databases upon navigation

mixonic:
@fancyremarker for you to review and merge.

We discussed having the apps and databases displayed reload each time a user navigates. This implements that behavior.

### DIFF
--- a/app/apps/route.js
+++ b/app/apps/route.js
@@ -3,6 +3,11 @@ import Ember from 'ember';
 export default Ember.Route.extend({
   model: function(){
     var stack = this.modelFor('stack');
-    return stack.get('apps');
+    var apps = stack.get('apps');
+    if (apps.get('isFulfilled')) {
+      return apps.reload();
+    } else {
+      return apps;
+    }
   }
 });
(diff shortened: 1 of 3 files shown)


app/apps/route.js
```diff
@@ -3,6 +3,11 @@ import Ember from 'ember';
 export default Ember.Route.extend({
   model: function(){
     var stack = this.modelFor('stack');
-    return stack.get('apps');
+    var apps = stack.get('apps');
+    if (apps.get('isFulfilled')) {
```
fancyremarker: What's this `isFulfilled` property?

fancyremarker:
:+1:

mixonic:
Great, merging. Fwiw "@fancyremarker for you to review and merge." means I absolutely leave it in the hands of the reviewing developer to click the green button. Saves us a little back-and-forth overhead :-)

app/apps/route.js
sandersonet: This gets the state of the app's `ArrayController` promise.  `isFulfilled` being true indicates the promise is resolved and apps already live on the client, thus the need to `reload()` them to get latest.   Any other states of the promise would indicate the request is still in flight, so it can be returned assuming it's the latest data.  (I had to look this up to figure it out, so correct me if I am wrong @mixonic)

mixonic: @sandersonet you are correct! `apps` is a promise proxy, and here we are inspecting the state of the promise to decide if it needs reloading or is already in flight.

Issue #87, opened 2015-02-16
audit history dates should be formatted as UTC timestamps


bantic:
The app activity and database activity pages should show operation `createdAt` dates as UTC timestamps in the format "monthName monthDay, fullYear HH:MMam/pm UTC":

![image](https://cloud.githubusercontent.com/assets/2023/6219472/edcbb192-b5f7-11e4-827e-87eb8e06af6f.png)

Issue #86, opened 2015-02-16
app show page


bantic:
from #51 

Show last deploy and current git ref on app show page.

![image](https://cloud.githubusercontent.com/assets/2023/6218881/fdfb8896-b5f0-11e4-9de4-e1f0a41596a7.png)

sandersonet:
@bantic, we actually don't have a Users#show page yet, so the user's name doesn't need to be a link.

bantic:
@sandersonet yeah, was just about to ask that. thanks :)

bantic:
@sandersonet the current git ref would be the one matching app.currentImage.gitRef, right? Not app.lastDeployOperation.gitRef? (Or are they the same?)

sandersonet:
You are correct, `app.currentImage.gitRef` is the correct ref.  Currently `app.lastDeployOperation` does not verify that the last deploy actually succeeded, so it's possible that Git ref isn't actually what's currently deployed.  We could probably add an icon or color indicator to show if that Last Deployed By was a success or failure.  (I can add that easily later on)

PR #88, opened 2015-02-16
Show last deploy operation user and time, current git ref on app show

### LINKED ISSUE #73: Properly handle app.lastDeployOperation and app.currentImage
These are of type operation and image, respectively, and are embedded by HAL.
The HAL adapter needs to be updated to handle embedded properties that have a different name (eg 'last_deploy_operation') than their type (eg 'operation').

See https://github.com/201-created/ember-data-hal-9000/issues/7

bantic:
* add `format-date` helper
  * add `iso-8601-timestamp` transform
  * add `image` model
  * DRY common model dependencies for unit tests
  * test that an app's `currentImage` and `lastDeployOperation` can be sideloaded

@mixonic for you
fixes #73 #86


bantic:
looks like 

image

bantic:
also format utc timestamps on app/db activity pages. fixes #87.

looks like

image

app/helpers/format-date.js
```diff
@@ -0,0 +1,14 @@
...
+import { monthNames } from 'diesel/utils/dates';
+
+export function formatDate(date) {
+  if (!date) { return; }
+
+  Ember.assert('format-date must be called with an instanceof Date',
+               date instanceof Date);
+
+  let monthName = monthNames[ date.getMonth() ];
+  return `${monthName} ${date.getDate()}, ${date.getFullYear()}`;
```
mixonic: @sandersonet @fancyremarker check out this use of ES2015 string interpolation. These are also multi-line capable.

tests/unit/models/operation-test.js
```diff
@@ -4,22 +4,13 @@ import {
 } from 'ember-qunit';
 import { stubRequest } from "../../helpers/fake-server";
 import Ember from 'ember';
+import modelDeps from '../../support/common-model-dependencies';
```
mixonic: Might want to refactor these to be called `stackChildModelDeps` or something later. `common` implies common to all models, and these are really constrained to children of stack.

PR #90, opened 2015-02-17
Handle input

mixonic:
Add dashes to handle input. On app, db, and first-app app/db creation. Closes https://github.com/aptible/diesel.aptible.com/issues/89.

Also includes some cleanup for CSP headers and a reversion to Ember 1.10. Ember beta and canary seem to have an issue with pushing the cursor to the end of the line even when the value does not change. See https://github.com/emberjs/ember.js/issues/10481.

### DIFF
--- a/app/components/handle-input/component.js
+++ b/app/components/handle-input/component.js
@@ -1,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.TextField.extend({
+  _dasherizedValue: null,
+  value: Ember.computed(function(key, value){
+    if (arguments.length > 1) {
+      if (value && typeof value === 'string') {
+        this._dasherizedValue = value.dasherize();
+      } else {
+        this._dasherizedValue = value;
+      }
+      return this._dasherizedValue;
+    } else {
+      return this._dasherizedValue;
+    }
+  })
+});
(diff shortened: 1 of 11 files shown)


tests/unit/components/handle-input/component-test.js
```diff
@@ -0,0 +1,39 @@
...
+
+  assert.equal(input.val(), 'new-val');
+  assert.equal(component.get('value'), 'new-val');
+});
+
+test('setting value with spaces adds dashes', function(assert) {
+  assert.expect(2);
+
+  var component = this.subject();
+  var element = this.append();
```
bantic: `append` is deprecated now. I'll rename this after merging.

app/components/handle-input/component.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.TextField.extend({
+  _dasherizedValue: null,
+  value: Ember.computed(function(key, value){
```
bantic: cool — I wonder if it might be helpful to add some of the [handle validations](https://github.com/aptible/api.aptible.com/blob/master/app/models/handle.rb#L4) from the API here in addition to turning " " to "-".

PR #91, opened 2015-02-17
show a checklist for a non-deployed app

### LINKED ISSUE #78: empty app view (screen 203)
from #51.

The empty app view screen shows info on how to deploy an app.
Should look like:

image

bantic:
@mixonic, for you

  * Adds 'pending' status type, `hasBeenDeployed` CP for app model
  * display checklist on app show page if app has not been deployed

fixes #78 

Also:

  * rename some acceptance test files to put 'app' related tests together

This makes it so that a never-been-deployed app is shown a screen with a checklist that looks like this:
image

or, if the user has no ssh keys:

image

This is ready to merge, but it will not work properly until the API returns a 'pending' status for never-been-successfully-deployed apps. I'm making a PR for that now.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -20,6 +20,10 @@ module.exports = function(environment) {
 
     segmentioKey: 'qxwhj8ys4t',
 
+    externalUrls: {
+      gettingStartedDocs: 'https://support.aptible.com/hc/en-us/articles/202638630-Deploying-your-first-app'
+    },
+
     EmberENV: {
       FEATURES: {
         // Here you can enable experimental features on an ember canary build
--- a/app/app/deploy/controller.js
+++ b/app/app/deploy/controller.js
@@ -1,0 +1,2 @@
+import Ember from 'ember';
+export default Ember.Controller.extend({});
--- a/app/app/deploy/route.js
+++ b/app/app/deploy/route.js
@@ -1,0 +1,22 @@
+import Ember from "ember";
+import config from '../../config/environment';
+
+let gettingStartedDocs = config.externalUrls.gettingStartedDocs;
+
+export default Ember.Route.extend({
+  model: function(){
+    let app = this.modelFor('app');
+    let currentUser = this.session.get('currentUser');
+
+    return Ember.RSVP.hash({
+      app: app,
+      sshKeys: currentUser.get('sshKeys')
+    });
+  },
+
+  setupController: function(controller, model){
+    controller.set('model', model.app);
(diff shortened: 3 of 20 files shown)


bantic:
After merging you'll want to reinstall aptible-sass to pick up the style changes from https://github.com/aptible/aptible-sass/pull/69

bantic:
Complementary PRs:

  * Sweetness https://github.com/aptible/sweetness/pull/264
  * API https://github.com/aptible/api.aptible.com/pull/164

config/environment.js
```diff
@@ -20,6 +20,10 @@ module.exports = function(environment) {
 
     segmentioKey: 'qxwhj8ys4t',
 
+    externalUrls: {
+      gettingStartedDocs: 'https://support.aptible.com/hc/en-us/articles/202638630-Deploying-your-first-app'
+    },
```
mixonic: If it is expected that we will have a number of these, we should consider just putting them in a `utils` file. We have a configurable list of `aptibleHosts` already, and the rest of the path seems unlikely to change from env to env.

Data in the `config/environment.js` is sent to the client at least twice, as we've discovered. Keeping it slim should be on our minds.

mixonic:
This looks really great @sandersonet @bantic!

sandersonet:
@fancyremarker @bantic I thought more about how we determine a 'successful' deployment and have some questions.  Right now, it's possible to successfully deploy an app without actually creating any services. For example, by deploying with either an empty or malformed Procfile, the deployment will succeed but no services will be created.

I think it's important that users know that services need to exist in order to have a working app.  Because of this, I propose we either:
a) Update sweetness to error on malformed or empty Procfile.  We do error if the Procfile is missing entirely.
b) Still show the deployment steps if the app is `provisioned`, but lacks any services.
c) Any other instance where a deployment succeeds, but services are not created?

Also, @fancyremarker, I actually don't know at what interval you and Aaron update sweetness on all of our stacks.  Will aptible/sweetness#264 get rolled out within the next two weeks?  The hope was to have diesel on production within that timeframe.

fancyremarker:
> a) Update sweetness to error on malformed or empty Procfile. We do error if the Procfile is missing entirely.

^^ This is definitely what we want to do. If that's not the case right now, we should open a Sweetness issue.

Re: aptible/sweetness#264, that will get deployed on next Sweetness deploy (can happen at any point), and doesn't require any of the stack migrations from v1—v2 to be completed (that's the part which will happen on a more delayed schedule).

PR #99, opened 2015-02-18
Deprovision databases

mixonic:
@bantic please review and merge when appropriate.

Quickly nips off database deprovisioning from https://github.com/aptible/diesel.aptible.com/issues/51. There is definitely duplicated code between this an apps, but I'd like to defer getting eager about DRY until we are further along.

### DIFF
--- a/app/database/deprovision/template.hbs
+++ b/app/database/deprovision/template.hbs
@@ -1,0 +1,35 @@
+<div class="row">
+  <div class="col-xs-9">
+    <div class="panel panel-default panel-subdued panel-danger">
+      <div class="panel-heading">
+        <h3>Permanently deprovision {{model.handle}} </h3>
+      </div>
+      <div class="panel-body">
+        <form role="form">
+          {{#if error}}
+            <div class="alert alert-warning fade in">
+              <a class="close" data-dismiss="alert">×</a>
+              <p>
+                There was a problem deprovisioning your database: {{error.responseJSON.message}}
+              </p>
+            </div>
+          {{/if}}
+
+          <p>
+            This action CANNOT be undone. This will permanently deprovision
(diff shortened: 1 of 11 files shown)


app/database/deprovision/template.hbs
```diff
@@ -0,0 +1,35 @@
...
+      <div class="panel-heading">
+        <h3>Permanently deprovision {{model.handle}} </h3>
+      </div>
+      <div class="panel-body">
+        <form role="form">
+          {{#if error}}
+            <div class="alert alert-warning fade in">
+              <a class="close" data-dismiss="alert">×</a>
+              <p>
+                There was a problem deprovisioning your database: {{error.responseJSON.message}}
```
bantic: We should eventually settle on some sort of general strategy for displaying errors. Errors that don't have this shape (like a syntax error or some generic 500 response from the server) would end up looking funny.

tests/helpers/aptible-helpers.js
```diff
@@ -97,6 +97,15 @@ Ember.Test.registerHelper('stubApp', function(app, appData){
   });
 });
 
+Ember.Test.registerHelper('stubDatabase', function(app, databaseData){
+  var id = databaseData.id;
+  if (!id) { throw new Error('cannot stub database without id'); }
+
+  stubRequest('get', '/databases/' + id, function(request){
```
bantic: this could be `stubRequest('get', '/databases/:db_id/', function(request){`, and then the response could dynamically fill in the id from `request.params.db_id`. But stubbing *any* db get call to return `databaseData` isn't the intent of this helper, so I think it makes sense to hard-code to the given id.

bantic:
awesome

Issue #101, opened 2015-02-18
use "deprovision" operation to remove log drains


bantic:
After discussion with @fancyremarker we shouldn't `DELETE /log_drains/:log_drain_id` from diesel directly. Better to use a deprovision operation like we do when deprovisioning apps and dbs. The UI may need to change to avoid displaying deprovisioned log drains (treating them as though they are deleted).

refs #98

bantic:
@fancyremarker Is it still ok to `POST /log_drains` from diesel to create a new log drain?

aaw:
Out of curiosity, what's the rationale for a separate deprovision job versus allowing Diesel to update the state of the log drains and then just run a configure log drain job?

bantic:
I'm not sure. Are you saying diesel would change status of the log drain to "deprovisioned" and at the same time create an operation for that log drain of type "configure"? I'm not sure what the configure job does but that may achieve the same effect as a separate deprovision job. I guess there may be potential problems having it be a two-phase process (change log drain status, then create deprovision op) from the client-side because only one of the two may succeed, leaving the log drain in a dangling state?

fancyremarker:
@aaw: I think a `deprovision` job is appropriate for the use case of removing a drain entirely. (For example, a user provisions a Papertrail drain, then adds an Elasticsearch drain, then wishes to remove the Papertrail drain.)

Does that sound right to you?

aaw:
Sorry, not trying to start a centithread - I'm just interested in understanding why Diesel shouldn't `DELETE /log_drains/:log_drain_id`, since that probably means that I'm approaching this the wrong way from Sweetness.

Some more background: on the Sweetness side, log drains are implemented with one logstash instance per account. Several forwarders send logs to the account's logstash instance and that instance forwards the logs to their final destination (syslog, elasticsearch, etc.) Any change of log drains needs to be followed by a "re-configure" of logging on the Sweetness side, which means a new logstash config is generated from the current state of all log drains and logstash is restarted. Right now, we have a ConfigureLogDrain job in sweetness that does just that. So whether you add a log drain, delete a log drain, or modify a log drain, it has to be followed by a ConfigureLogDrain in Sweetness to take effect.

[...]

PR #102, opened 2015-02-18
Display, create Log drains

### LINKED ISSUE #98: logging pages (screen 210, 211)
from #51 

  - [x] log drains index page
  - [x] log drains create page
  - ~~[ ] can delete log drain from index page~~ (see #101)
  - [x] make styles match up
  - [x] DRY similarities in tests
  - [x] display log drain info in stack/-header.hbs (show first log drain's info)

Add a "logging" tab for a page that shows a list of logs, like so:
image

And a "new log" page that allows a user to create a new log drain, like so:
image

bantic:
fixes #98

@mixonic, for you

Displays a list of log drains on the stacks/:id/logging page, with a button to add a new log drain.

Deleting a log drain is not supported yet (see #101).

The stack header shows info about the first log drain. That may need to change later based on feedback.

Looks like:

image

and

image


bantic:
Hold off on this — the url for POSTing to create a log drain is different than GETting one. I need to make some updates.

bantic:
Ready to go, again. Added 'handle' and 'drainType' attrs to logDrain to pass server-side validations. Form now looks like:

image

app/stack/log-drains/new/route.js
```diff
@@ -0,0 +1,29 @@
...
+    createLog: function(log){
+      let route = this;
+
+      log.set('stack', this.modelFor('stack'));
+
+      log.save().then(function(){
+        route.transitionTo('stack.log-drains.index');
+      }).catch(function(e){
+        if (!(e instanceof DS.InvalidError)) {
+          return e;
```
mixonic: `return e` will squash the failed promise. Is that the intent?

bantic: good catch. fixing this...

PR #103, opened 2015-02-18
Change replaceWith calls to transitionTo to fix back button issues

### LINKED ISSUE #95: Back button is broken
If you navigate away from Stack#show and then click the back button, you are not returned to Stack#show.

bantic:
@mixonic, for you. I'm not sure why we had as many `replaceWith` calls as
we did. I've hand-tested for deleterious effects of making these changes
and have not found any.

fixes #95

### DIFF
--- a/app/app/index/route.js
+++ b/app/app/index/route.js
@@ -5,9 +5,9 @@ export default Ember.Route.extend({
     let app = this.modelFor('app');
 
     if (app.get('hasBeenDeployed')) {
-      this.replaceWith('app.services');
+      this.transitionTo('app.services');
     } else {
-      this.replaceWith('app.deploy');
+      this.transitionTo('app.deploy');
     }
   }
 });
--- a/app/database/index/route.js
+++ b/app/database/index/route.js
@@ -2,6 +2,6 @@ import Ember from 'ember';
 
 export default Ember.Route.extend({
   redirect: function(){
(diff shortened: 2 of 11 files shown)


mixonic:
transitionTo makes a history entry. if you are redirected and click back, you often don't want to return to the redirect itself. Instead the redirect is mean to rewrite where you currently are. Obviously this isn't true 100% of the time, but I think in many of these cases it is true.

For instance the database index route. There is no reason I would want a click of the back button to return me to that route (which does nothing but redirect).

bantic:
@mixonic if you change the `transitionTo` to a `replaceWith` in the database/index/route.js you get this behavior (for me, in Chrome):

  * Click "apps" tab
  * click "databases" tab
  * click a database
  * back button
  * end up erroneously on "apps" page

Changing to `transitionTo` fixes this for me.

Screencap:
db

mixonic:
Ok. My understanding must be flawed? I would like to know what is actually happening here. I don't doubt these changes improved some specific behavior, I was wondering if the change was warranted in so many places.

I'll pull it down and figure out what is happening then report back.

bantic:
let me know what you find, or if you want to pair. It doesn't fit my mental model of how `replaceWith` works, either.

mixonic:
Let's review this in the AM.

PR #104, opened 2015-02-18
Test that cancel/click exit from app/db create rolls back model

### LINKED ISSUE #80: Exiting a view should effectively cancel unsaved changes
When creating a new App (or DB),  a new record is created immediately when the route loads.  If I then browse to another route and come back to apps/index, I see my unsaved app in the collection of other apps.  

Unsaved resources should be immediately destroyed when the route exits and should not be accessible from any index pages.

bantic:
fixes #80

@sandersonet, for you. This uses the `willTransition` hook to roll
back the app/db from the app/db 'new' pages.

Cleaned up the app/db acceptance tests a little and added two new tests
that after transitioning away from an in-progress app/db creation that
app/db doesn't show up on the index page.

cc @mixonic

### DIFF
--- a/app/apps/new/template.hbs
+++ b/app/apps/new/template.hbs
@@ -20,7 +20,7 @@
               App handle
               <span class="label-helper">Lowercase alphanumerics, periods, and dashes only</span>
             </label>
-            {{handle-input class="form-control app-handle" value=model.handle}}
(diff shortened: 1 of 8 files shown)


mixonic:
:+1:

bantic:
I will rebase this in a moment

bantic:
rebased and ready to merge

PR #105, opened 2015-02-18
Stacks index

### LINKED ISSUE #92: Stacks index page
After some thought, I think having an actual stacks index page is a better UX than using a dropdown for stack switching.  

205d stacks

sandersonet:
For you to review @mixonic 

This PR updates the default route to be a stacks index page that shows a listing of all stacks a user has access to, ordered by organization.  If a user only has access to a single stack, then that stack's show page is the default route.

Fixes #92


app/templates/stack/-link.hbs
```diff
@@ -0,0 +1,14 @@
+{{#link-to 'stack' stack class="panel-link"}}
+  <div class="panel panel-default panel-subdued stack resource-list-item">
+    <div class="panel-heading">
+      <h5>{{stack.name}}</h5>
+    </div>
+    <div class="panel-body">
+      {{#if stack.allowPHI}}
+        <span class="success">PHI Ready</span>
+      {{else}}
+        <span class="success">Not PHI Ready</span>
```
bantic: this should probably be a `class="danger"`

sandersonet: Thanks!

app/components/stacks-by-organization/component.js
```diff
@@ -0,0 +1,24 @@
...
+      if(!groups.findBy('organizationId', currentOrganizationId)) {
+        groups.addObject(Ember.Object.create({
+          organizationId: currentOrganizationId,
+          organization: currentOrganization,
+          stacks: Ember.A([])
+        }));
+      }
+
+      groups.findBy('organizationId', currentOrganizationId).get('stacks').addObject(stack);
+    });
```
mixonic: As written, this logic will always traverse `groups` twice for each stack. It can be made more performance by using a reference to the found organization:

```js
  organizationGroups: function() {
    var groups = [];
    this.get('stacks').forEach(function(stack) {
      var currentOrganizationId = stack.get('_data.links.organization');
      var currentOrganization = stack.get('organization');
      var group = groups.findBy('organizationId',currentOrganizationId);
      if (!group) {
        group = {
          organizationId: currentOrganizationId,
          organization: currentOrganization,
          stacks: []
        };
        groups.push(group);
      }
      group.stacks.push(stack);
    });
    return groups;
  }).property('stacks.[]')
```

This version will only traverse groups once for each stack.

Note that I'm not using Ember's object and array classes. You really only need Ember objects when you want something that is **observable**. In this case, the data structure is only created once then updated. Hence it does not need to use Ember objects.

Because Ember ships with prototype extensions enabled by default (this may change in 2.0), you can still use `findBy` on the native array.

Also note that I've make the property dependent upon `stacks.[]`. The `.[]` means "observe the contents of this array". Without the brackets, only the array itself would be observed, and since arrays are references (the array itself is the same no matter how the contents change), your property would not become invalid and re-compute if new records were added.

mixonic:
It took me a while to understand what was going on here, but I see it now. `stacks` is now a top level route on its own with content specific to itself. `stack` and its child routes do not share layout or async dependencies (loaded models) with `stacks`, and thus is extracted to be a top level item all its own.

I think this could be more simply achieved by moving some of what we now have in `stacks` into `stacks.index`. That would have avoided many of the changes. That said, it doesn't look like pretty much all of `stacks` would move into `stacks.index` which leaves one questioning why it should be left around at all.

In summary I'm +1, pending the changes to `organizationGroups` to be less of a performance bump.

app/components/stacks-by-organization/component.js
mixonic: Oh, also, is it possible to read the `self` link from `organization` instead of peeking into `_data`? That is definitely a private API and not very stable to build upon.

@bantic is there a good way to read an arbitrary link URL or object via meta or something?

bantic: Unfortunately not. For a non-collection resource (i.e. a singular one) we don't store the links in meta.

sandersonet:
@mixonic Thanks for the feedback.  I've made the recommended improvements along with removing the `stacks.index` route and template, rebasing, and fixing specs.

Issue #94, opened 2015-02-17
Update document page title on route transition


sandersonet:
- [x] Index => 'Aptible Dashboard'
- [x] Stack#show#apps => "${stack.handle} Apps - ${organization.name}"
- [x] Stack#show#apps#new => "Create an App - ${stack.handle}"
- [x] Stack#show#databases => "${stack.handle} Databases - ${organization.name}"
- [x] Stack#show#databases#new => "Create a Database - ${stack.handle}"
- [x] Stack#show#logging => "${stack.handle} Logging - ${organization.name}"
- [x] App#show => "${app.handle} - ${stack.handle}" (note: this route currently immediately transitions to App#show#services, so this title would never be seen)
- [x] App#show#services => "${app.handle} Services - ${stack.handle}"
- [x] App#show#domains => "${app.handle} Domains - ${stack.handle}"
- [x] App#show#domains#new => "Add a domain - ${app.handle} - ${stack.handle}"
- ~~[ ] App#show#domains#edit => "Edit ${domain.virtual_domain} - ${app.handle}"~~ (see #112)
- [x] App#show#activity => "${app.handle} Activity - ${stack.handle}"
- [x] App#show#deprovision => "Deprovision ${app.handle} - ${stack.handle}"
- [x] Database#show#activity => "${database.handle} Activity - ${stack.handle}"
- [x] Database#show#deprovision => "Deprovision ${database.handle} - ${stack.handle}"

mixonic:
Assigning this to myself.

PR #106, opened 2015-02-18
Introduce support for titles

mixonic:
Fixes much but not all of #94 (some of those screens are still in progress). I've added a checklist there.

This introduces support for titles. Each route can have a `title` property. After transition, the route tree will be walked upward until a title string or function is found. This function is then called.

There is also support for building URLs with tokens. I'm going to look at a refactor using this logic now that the tests are in place.

WIP, please do not merge.


mixonic:

app/utils/title-route-extensions.js
```diff
@@ -0,0 +1,81 @@
...
+  // Provided by Ember
+  _actions: {
+    collectTitleTokens: function(tokens) {
+      var titleToken = this.titleToken;
+      if (typeof this.titleToken === 'function') {
+        titleToken = this.titleToken(this.currentModel);
+      }
+
+      if (Ember.isArray(titleToken)) {
+        tokens.unshift.apply(this, titleToken);
```
bantic: is there a performance benefit to doing it this way instead of `tokens = titleToken.concat(tokens)` ?

bantic: and I may be reading this wrong but wouldn't it be `tokens.unshift.apply(tokens, titleToken)`?

bantic:
It's funny how cool watching a page's title change dynamically can be. Neat to see this in action. The use of 'token' throws me off a little — I would think that would be named like 'breadCrumbs' or something. I wonder if this is a more complicated/comprehensive solution than we strictly need, but it is likely be better suited for future use cases than a simpler system.

I had a question about `tokens.unshift.apply` above — after double-checking me on that, please merge at your leisure.

app/utils/title-route-extensions.js
mixonic: Using `concat` would allocate a second array. I agree this should be `apply(tokens` though, confirming.

PR #109, opened 2015-02-19
Adds organization settings links to user nav

### LINKED ISSUE #93: Multiple organization UX
Rather than listing all organizations across the top, I think we can fit them into the user dropdown on top.  For now, these will just link to the existing dashboard's organizations#show pages. 

205e organizations

sandersonet:
Fixes #93 

This PR requires https://github.com/aptible/auth.aptible.com/pull/119 to work properly.

For you @mixonic.

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -17,12 +17,35 @@ export default DS.Model.extend({
   // relationships
   token: DS.belongsTo('token', {async: true}),
   roles: DS.hasMany('role', {async:true}),
-  organizations: DS.hasMany('organization', {async:true}),
+  //organizations: DS.hasMany('organization', {async:true}),
   sshKeys: DS.hasMany('ssh-key', {async:true}),
 
   // check ability, returns a promise
   // e.g.: user.can('manage', stack).then(function(boolean){ ... });
   can: function(scope, stack){
     return can(this, scope, stack);
-  }
+  },
+
+  organizations: function() {
+    var organizations = [];
+
+    this.get('roles').forEach(function(role) {
+      var organizationId = role.get('_data.links.organization');
+      var organization = organizations.findBy('organizationId', organizationId);
+
+      if(!organization) {
+        organization = {
+          organizationId: organizationId,
+          organization: role.get('organization')
+        };
+        organizations.push(organization);
+      }
+    });
+
+    return organizations.map(function(organization) {
+      return organization.organization;
+    });
+
+  }.property('roles.[]')
+
 });
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -1,5 +1,6 @@
 import DS from 'ember-data';
 import can from "../utils/can";
+import Ember from 'ember';
 
 export default DS.Model.extend({
   email: DS.attr('string'),
@@ -17,7 +18,6 @@ export default DS.Model.extend({
   // relationships
   token: DS.belongsTo('token', {async: true}),
   roles: DS.hasMany('role', {async:true}),
-  //organizations: DS.hasMany('organization', {async:true}),
   sshKeys: DS.hasMany('ssh-key', {async:true}),
 
   // check ability, returns a promise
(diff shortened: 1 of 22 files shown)


sandersonet:
@mixonic aptible/auth.aptible.com#119 will cause issues by exposing organizations to unauthorized users, so I instead use a computed property on `models/user.js` that returns all unique organization through roles.  Let me know if there is a better way to do this.

app/models/user.js
```diff
@@ -17,12 +17,35 @@ export default DS.Model.extend({
   // relationships
   token: DS.belongsTo('token', {async: true}),
   roles: DS.hasMany('role', {async:true}),
-  organizations: DS.hasMany('organization', {async:true}),
+  //organizations: DS.hasMany('organization', {async:true}),
```
mixonic: Should this line just be removed?

app/models/user.js
```diff
@@ -17,12 +17,35 @@ export default DS.Model.extend({
...
+        };
+        organizations.push(organization);
+      }
+    });
+
+    return organizations.map(function(organization) {
+      return organization.organization;
+    });
+
+  }.property('roles.[]')
```
mixonic: I would simplify:

```javascript
  organizations: function() {
    return this.get('roles').mapBy('organization').uniq();
  }.property('roles.@each.organization')
```

Some references, I know this uses some slick trickery:

* [uniq](http://emberjs.com/api/classes/Ember.Enumerable.html#method_uniq)
* [mapBy](http://emberjs.com/api/classes/Ember.Enumerable.html#method_mapBy)
* [@each](http://emberjs.com/guides/object-model/computed-properties-and-aggregate-data/)

It should have the same effect though, and perform maybe a little better.

app/templates/-current-user-header.hbs
```diff
@@ -6,34 +6,18 @@
...
           {{#dashboard-dropdown title=session.currentUser.name defaultPath="index"}}
 
-            {{#dashboard-dropdown-item}}
-              {{link-to "Dashboard" "index"}}
-            {{/dashboard-dropdown-item}}
-
-            {{#dashboard-dropdown-item}}
-              {{#link-to "settings"}}Settings{{/link-to}}
-            {{/dashboard-dropdown-item}}
+            {{#each organization in session.currentUser.organizations}}
```
mixonic: The block params style is preferred:

```handlebars
{{#each session.currentUser.organizations as |organization|}}
  {{dashboard-dropdown-organization-menu organization=organization}}
  {{dashboard-dropdown-divider}}
{{/each}}
```

Though not required. The `in` is basically remapped at compile time to what I wrote out here.

mixonic:
Some failing tests and I left some comments. I an generall :+1: with the rest though!

app/models/user.js
sandersonet: Thanks Matt, this is an awesome simplification.

`uniq()` doesn't seem to work though.  According to the docs:

> This only works on primitive data types, e.g. Strings, Numbers, etc.

Is there a way to get it to work with models?

mixonic: Shoot, you are right. Looks like it needs to be a bit more complex then:

```javascript
var organizations = [];
this.get('roles').forEeach(function(role){
  var organization = role.get('organization');
  if (organizations.indexOf(organization) === -1) {
    organizations.push(organization);
  }
});
return organizations;
```

sandersonet:
cc @mixonic 

I updated the `stacks` route to include all of a user's organizations and use those to loop through and generate the `stacks` page.

app/templates/-current-user-header.hbs
```diff
@@ -6,34 +6,18 @@
...
           {{#dashboard-dropdown title=session.currentUser.name defaultPath="index"}}
 
-            {{#dashboard-dropdown-item}}
-              {{link-to "Dashboard" "index"}}
-            {{/dashboard-dropdown-item}}
-
-            {{#dashboard-dropdown-item}}
-              {{#link-to "settings"}}Settings{{/link-to}}
-            {{/dashboard-dropdown-item}}
+            {{#each session.currentUser.organizations as |organization|}}
```
mixonic: This is where I might expect that organizations (already fetch by the route) is referenced.

mixonic:
@sandersonet I think there are some opportunities to reduce the number of organizations queries by fetching them from a top-level route. It feels like we are implying an `organizations` route at the top level that always fetches.

But let us keep moving forward. Excellent work here!

PR #111, opened 2015-02-20
Redirect empty index pages to new page, fixes #100

### LINKED ISSUE #100: Consider a design for empty index pages
- Apps
- Databases
- Log drains

sandersonet:
@mixonic 

This PR allows us to skip showing empty index pages and instead redirect the user directly to the new resource route.  Fixes #100

### DIFF
--- a/app/apps/route.js
+++ b/app/apps/route.js
@@ -13,5 +13,11 @@ export default Ember.Route.extend({
     } else {
       return apps;
     }
+  },
+  redirect: function() {
+    var stack = this.modelFor('stack');
+    if(stack.get('apps.length') === 0) {
+      this.replaceWith('apps.new', this.modelFor('stack'));
+    }
   }
 });
--- a/app/apps/route.js
+++ b/app/apps/route.js
@@ -13,10 +13,5 @@ export default Ember.Route.extend({
     } else {
       return apps;
     }
-  },
-  redirect: function(model) {
-    if(model.get('length') === 0) {
-      this.replaceWith('apps.new', this.modelFor('stack'));
-    }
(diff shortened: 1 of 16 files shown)


sandersonet:
Hold up, just realized this doesn't work exactly right on vhosts.

app/apps/route.js
```diff
@@ -13,5 +13,11 @@ export default Ember.Route.extend({
     } else {
       return apps;
     }
+  },
+  redirect: function() {
+    var stack = this.modelFor('stack');
+    if(stack.get('apps.length') === 0) {
+      this.replaceWith('apps.new', this.modelFor('stack'));
+    }
```
bantic: The `redirect` hook is passed the resolved `model` (the array of apps), so I think you can do:
```
redirect: function(model){
  if (model.get('length') === 0) {
    this.transitionTo('stack.apps.new', this.modelFor('stack');
  }
}
```

app/apps/route.js
```diff
@@ -13,5 +13,10 @@ export default Ember.Route.extend({
     } else {
       return apps;
     }
+  },
+  redirect: function(model) {
+    if(model.get('length') === 0) {
+      this.replaceWith('apps.new', this.modelFor('stack'));
+    }
```
mixonic: perhaps this and the databases routes should also only be redirecting in `index`.

And I'd like to make these `transitionTo` instead of `replaceWith`.

mixonic:
Great testing.

sandersonet:
Thanks @bantic and @mixonic, I've made your suggested changes.

PR #118, opened 2015-02-20
Defer readiness when checking for `verification_code` param

### LINKED ISSUE #115: deferReadiness when redirecting due to old verification urls

bantic:
Only advance readiness if the code is not found.
This prevents a routing error when ember tries to recognize the
"/verify?verification_code" url.

fixes #115

@mixonic for you


mixonic:
@bantic I'm actually somewhat wary of putting deferral stuff into initializers. IMO all initializers should be doing is setting up registrations and injections. If they start doing more they may look up an object from the container before all reg/inject are present. This way leads the pain.

But, it works.

PR #121, opened 2015-02-23
App#Vhosts#Edit Page

### LINKED ISSUE #112: App#Domain#Edit
There's an 'edit' link on a vhost (domain) show page, but no route/template for editing yet.

 - [x] Edit page updates title to "Edit ${domain.virtual_domain} - ${app.handle}" (from #94)

  - [x] style to match other pages
  - [x] create "reprovision" operation after editing

@sandersonet Any important difference between an "add domain" page (shown below) and an "edit domain" page? Anything that should not be changeable when editing an existing domain?

image

bantic:
* Add tests for viewing edit page, save with and without error
  * Ensure that updating a vhost PUTs to correct url
  * Create a reprovision op when updating the VHost succeeds

Fixes #112

### DIFF
--- a/app/adapters/vhost.js
+++ b/app/adapters/vhost.js
@@ -1,8 +1,31 @@
 import ApplicationAdapter from './application';
 import buildURLWithPrefixMap from '../utils/build-url-with-prefix-map';
+import Ember from 'ember';
 
 export default ApplicationAdapter.extend({
   buildURL: buildURLWithPrefixMap({
     'services': 'service.id'
-  })
+  }),
+
+  // https://github.com/emberjs/data/blob/64ccb96673b2396dc16424b8ab1271afa43134ca/packages/ember-data/lib/adapters/rest_adapter.js#L549
+  // To update a VHost, the adapter should send a PUT request to /vhosts/:id,
+  // so we override updateRecord and remove the prepended /services/:service_id
+  // that `buildURL` adds
+  updateRecord: function(store, type, record){
+    var data = {};
+    var serializer = store.serializerFor(type.typeKey);
+
+    var snapshot = record._createSnapshot();
+    serializer.serializeIntoHash(data, type, snapshot);
+
+    var id = Ember.get(record, 'id');
+
+    var serviceId = Ember.get(record, 'service.id');
+    var url = this.buildURL(type.typeKey, id, record);
+    if (serviceId) {
+      url = url.replace("/services/" + serviceId, "");
+    }
+
+    return this.ajax(url, "PUT", { data: data });
+  }
 });
--- a/app/app/vhosts/edit/route.js
+++ b/app/app/vhosts/edit/route.js
@@ -1,0 +1,51 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default Ember.Route.extend({
+  title: function(){
+    let app = this.modelFor('app');
+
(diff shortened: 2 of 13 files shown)


bantic:
Looks like the "new" page, more or less:

image

bantic:
@mixonic for your review

app/adapters/vhost.js
```diff
@@ -1,8 +1,31 @@
...
+    var id = Ember.get(record, 'id');
+
+    var serviceId = Ember.get(record, 'service.id');
+    var url = this.buildURL(type.typeKey, id, record);
+    if (serviceId) {
+      url = url.replace("/services/" + serviceId, "");
+    }
+
+    return this.ajax(url, "PUT", { data: data });
+  }
```
mixonic: So, we do something similar with `delete`:

https://github.com/bantic/diesel.aptible.com/blob/app-domain-edit/app/adapters/application.js#L40

I think they should likely be a universal, application-adapter level change. All updates likely happen at a shallow namespace (as deletes do).

app/app/vhosts/edit/route.js
```diff
@@ -0,0 +1,51 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+export default Ember.Route.extend({
+  title: function(){
+    let app = this.modelFor('app');
+
+    return `Edit ${this.currentModel.get('virtualDomain')} - ${app.get('handle')}`;
```
mixonic: Should be able to:

```
`Edit ${this.currentModel.get('virtualDomain')}`
```

And the title logic in:

https://github.com/bantic/diesel.aptible.com/blob/app-domain-edit/app/app/route.js#L4

Will take care of the app handle.

app/app/vhosts/edit/route.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+      });
+    },
+
+    cancel: function(){
+      this.transitionTo('app.vhosts.index');
+    },
+
+    willTransition: function(){
+      this.currentModel.rollback();
```
mixonic: Should this only happen if dirty? I presume it has be no-op for you thus far.

app/app/vhosts/edit/route.js
bantic: True but if I understand it the title logic will also append the stack name, which we don't want.

app/app/vhosts/edit/route.js
bantic: yeah, it is a no-op if the model is not dirty.

app/app/vhosts/edit/route.js
mixonic: Ahha! You are correct. Excellent.

app/adapters/vhost.js
bantic: You may be right — I'm not sure how best to generalize this though, because not all of the models that we send updates on have anything prepended to them.

Looking through the test suite, the only PUTs we do at the moment are `/users/:user_id`, `/services/:service_id` and `/vhosts/:vhost_id`. The first two of those do not have anything prepended to their urls, but the latter does.

The only time we call `DELETE` in the test suite at the moment is for the ssh keys, so there would be no harm in generalizing the `destroyRecord` override (but also no gain)

mixonic: Sure :-p this is how the API is designed though. Deletes and updates are shallow, findalls and creates are nested. Let's just keep an eye open to make sure the logic isn't duplicated into another adapter later on.

PR #125, opened 2015-02-23
Add support for accepting invites

mixonic:

Invites can be accepted by signed in users. Users who are not signed in will be redirected to login/signup and redirected back once they
complete the process.

The invite page itself renders, then starts the AJAX request to claim the invitation automatically.

URLs such as:

* `http://localhost:4200/claim?verification_code=abc&invitation_id=59ef9caa-292c-42b0-b90f-000b7432b743`
* `http://localhost:4200/claim/59ef9caa-292c-42b0-b90f-000b7432b743/abc-verification-code`

Are supported. This is compatible with emails out there in the wild.

This PR introduces general support for redirecting to a path after signing in. See the `attemptedTransition` property.

### DIFF
--- a/app/claim/route.js
+++ b/app/claim/route.js
@@ -1,0 +1,31 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+  unauthenticatedRedirect: 'signup',
+
+  model: function(params){
+    return {
+      verificationCode: params.verification_code,
+      invitationId: params.invitation_id,
+      type: 'invitation'
+    };
+  },
+
+  actions: {
+    claim: function(){
+      var options = this.currentModel;
+      var verification = this.store.createRecord('verification', options);
+
+      verification.save().then( () => {
+        this.replaceWith('index');
+      }, () => {
+        this.controllerFor('claim').set('error', `
+          There was an error accepting this invitation. Perhaps
+          the verification code has expired?
+        `);
+      });
+    }
+  }
+
+
+});
--- a/app/claim/template.hbs
+++ b/app/claim/template.hbs
@@ -1,0 +1,26 @@
+{{#login-box}}
+  <div class="row">
+    <div class="col-xs-6 col-xs-offset-3">
+      <div class="panel panel-subdued panel-focused">
+        <div class="panel-heading">
+          <h3>Invitation Acceptance</h3>
+        </div>
+        <div class="panel-body">
+          {{#if error}}
+            {{#bs-alert alert="danger"}}
+              {{error}}
+            {{/bs-alert}}
+          {{/if}}
+          <p>
+            You are currently signed in as {{session.currentUser.email}}.
+          </p>
+          <form>
+            <button class="btn btn-primary btn-block btn-lg" {{action "claim"}}>
+              Accept organization invitation
+            </button>
+          </form>
+        </div>
+      </div>
+    </div>
+  </div>
+{{/login-box}}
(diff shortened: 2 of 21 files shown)


app/claim/route.js
```diff
@@ -0,0 +1,31 @@
...
+      var options = this.currentModel;
+      var verification = this.store.createRecord('verification', options);
+
+      verification.save().then( () => {
+        this.replaceWith('index');
+      }, () => {
+        this.controllerFor('claim').set('error', `
+          There was an error accepting this invitation. Perhaps
+          the verification code has expired?
+        `);
```
bantic: Doing this in `didTransition` allows the template to render — is that the reason for doing this in `didTransition`? I'm not clear on the pros/cons of async in this hook. Can the user get in a weird state by clicking away while this save is still `inFlight`?

app/claim/template.hbs
```diff
@@ -0,0 +1,22 @@
...
+    <div class="col-xs-6 col-xs-offset-3">
+      <div class="panel panel-subdued panel-focused">
+        <div class="panel-heading">
+          <h3>Invitation Verification</h3>
+        </div>
+        <div class="panel-body">
+          {{#if error}}
+            <div class="alert alert-danger fade in">
+              {{error}}
+            </div>
```
bantic: fwiw we now have `{{#bs-alert alert="danger"}}` for displaying bootstrappy alerts.

mixonic: Will definitely update before merge. Thanks.

tests/acceptance/claim-test.js
```diff
@@ -0,0 +1,128 @@
...
+  expect(3);
+  stubStacks(); // For loading index
+  stubOrganization();
+  stubOrganizations();
+  stubUser();
+  var verificationCode = 'some-code';
+  var invitationId = 'some-invitation';
+
+  stubRequest('post', '/verifications', function(request){
+    var params = JSON.parse(request.requestBody);
```
bantic: can do `this.json(request)`

app/claim/route.js
mixonic: There isn't anything to click on, the layout is akin to a login box.

The goal is to show some content on the page which the verification is happening. To show a box letting the user know something is going on.

As I've gone though the passwords flow, I started thinking about this UX more. IMO, at the least the user should see what their current email address is before they accept an invitation. This would help protect from automatic acceptance of an invite when a user clicks on a link.

In fact now that I think of it more, this is technically an XSS attack 0_o. You could make an arbitrary user accept an invite just by linking them, or even opening the URL in an iframe.

So, this flow should really be replaced with a button.  @sandersonet if that sounds good to you I can refactor this to have a box prompting the user with whatever easy data we already have for now. At least their current email.

bantic:
cool — I had a question about the impact of doing the async in the didTransition hook. I'm curious if that can be problematic, but I'm sure it's fine. This LGTM

app/claim/route.js
sandersonet: :+1:  to the button

mixonic:
Adds a button, addresses a few items caught by @bantic.

The screen for acceptance is very, very simple:



Significantly it does't have the organization name. This is passed with the URL from the email, but we should not be trusting that value. Instead, the API needs to be updated to allow us to easily fetch the organization for an invite. We can load that in the model hook and display the name here.

IMO there should also be a "this isn't me" button, or a "cancel" button.

I'd like to make those items a separate issue and keep moving forward though.

@bantic I think this is good to go.

bantic:
nice. I like the look of that button. So inviting. So accepting.

PR #127, opened 2015-02-23
Forgot Password

mixonic:
This is the UI and functionality for creating a password reset.

TODO:

* [x] Entering a new password. See https://github.com/aptible/auth.aptible.com/pull/121 for some API-side work.
* [ ] ~~Change the ugly `/password/resets/new` URL on the API to be something more RESTful.~~
* [x] ensure password resets are stored on the auth API

OK! This works!

Much flow. A new validation for password complexity. Much awesome.

### DIFF
--- a/app/adapters/password-reset-request.js
+++ b/app/adapters/password-reset-request.js
@@ -1,18 +1,18 @@
-import ApplicationAdapter from "./application";
+import AuthAdapter from "./auth";
 import Ember from "ember";
 
 export var auth = {};
 
-export default ApplicationAdapter.extend({
+export default AuthAdapter.extend({
 
   buildURL: function(){
     var url = [],
         host = Ember.get(this, 'host'),
         prefix = this.urlPrefix();
 
-    if (prefix) { url.unshift(prefix); }
+    if (prefix) { url.push(prefix); }
 
-    url.unshift('password', 'resets', 'new');
+    url.push('password/resets/new');
 
     url = url.join('/');
     if (!host && url) { url = '/' + url; }
--- a/app/adapters/password-reset-request.js
+++ b/app/adapters/password-reset-request.js
@@ -1,0 +1,23 @@
+import ApplicationAdapter from "./application";
+import Ember from "ember";
+
+export var auth = {};
+
+export default ApplicationAdapter.extend({
+
+  buildURL: function(type){
+    var url = [],
+        host = Ember.get(this, 'host'),
+        prefix = this.urlPrefix();
+
+    if (prefix) { url.unshift(prefix); }
+
+    url.unshift('password', 'resets', 'new');
+
+    url = url.join('/');
+    if (!host && url) { url = '/' + url; }
+
+    return url;
+  }
+
+});
--- a/app/adapters/password-reset-request.js
+++ b/app/adapters/password-reset-request.js
@@ -5,7 +5,7 @@ export var auth = {};
 
 export default ApplicationAdapter.extend({
 
-  buildURL: function(type){
+  buildURL: function(){
     var url = [],
         host = Ember.get(this, 'host'),
         prefix = this.urlPrefix();
--- a/app/mixins/routes/disallow-authenticated.js
+++ b/app/mixins/routes/disallow-authenticated.js
@@ -1,0 +1,19 @@
+import Ember from "ember";
+
+// If a user is authenticated, redirect them to `/`. If they
+// are not authenticated, allow them to continue.
+export default Ember.Mixin.create({
+  requireAuthentication: false,
+  beforeModel: function(){
+    return new Ember.RSVP.Promise((resolve, reject) => {
+      if (this.session.get('isAuthenticated')) {
+        reject();
+      } else {
+        this.session.fetch('aptible').then(reject, resolve);
+      }
+    }).catch(() => {
+      this.transitionTo('index');
+    });
+  }
+
+});
--- a/app/password/new/controller.js
+++ b/app/password/new/controller.js
@@ -6,7 +6,9 @@ export default Ember.Controller.extend(EmberValidations.Mixin, {
 
   validations: {
     'model.password': {
-      confirmation: { message: 'does not match password' }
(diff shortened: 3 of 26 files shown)


mixonic:
@bantic ok, this is ready for Real Review and merge (if ready)

app/adapters/password-reset-request.js
```diff
@@ -0,0 +1,23 @@
...
+        prefix = this.urlPrefix();
+
+    if (prefix) { url.unshift(prefix); }
+
+    url.push('password', 'resets', 'new');
+
+    url = url.join('/');
+    if (!host && url) { url = '/' + url; }
+
+    return url;
```
bantic: to support the custom '/password/resets/new' url?

mixonic: correct.

mixonic:
Lame. RecordIt silently truncated my gif animation.

app/mixins/routes/disallow-authenticated.js
```diff
@@ -0,0 +1,17 @@
...
+  beforeModel: function(){
+    return new Ember.RSVP.Promise((resolve, reject) => {
+      if (this.session.get('isAuthenticated')) {
+        resolve();
+      } else {
+        this.session.fetch('aptible').then(resolve, reject);
+      }
+    }).then(() => {
+      this.transitionTo('index');
+    }, function(){});
```
bantic: The idea that rejections are no-ops but resolutions, which indicate that the user is authenticated, get redirected to `/`? I would have expected a no-op successful `resolve` when the user is in the desired state (not authenticated), and a `.catch` at the end of the promise that transitions the authenticated user to '/'.

app/password/new/controller.js
```diff
@@ -0,0 +1,31 @@
...
+
+    save: function(model){
+      var hasSubmitted = this.get('hasSubmitted');
+      if (!hasSubmitted) {
+        this.set('hasSubmitted', true);
+      }
+      this.validate().then(() => {
+        this.get('target').send('save', model);
+      }).catch(function(){
+        // Silence the validation exception
```
bantic: would this swallow an error if the server responds with a 500?

mixonic: no. That error would occur in a different frame of JavaScript and not be caught here.

mixonic:
@bantic I've reversed and commented on the disallow mixin, cleaned up the buildURL to be a bit more direct. Please merge or drop me some more feedback if ya got it.

bantic:
:+1: thanks, will do in a moment

app/mixins/routes/disallow-authenticated.js
```diff
@@ -0,0 +1,19 @@
...
+// If a user is authenticated, redirect them to `/`. If they
+// are not authenticated, allow them to continue.
+export default Ember.Mixin.create({
+  requireAuthentication: false,
+  beforeModel: function(){
+    return new Ember.RSVP.Promise((resolve, reject) => {
+      if (this.session.get('isAuthenticated')) {
+        reject();
+      } else {
+        this.session.fetch('aptible').then(reject, resolve);
```
bantic: this should be `return`'d so that the `catch` below gets it?

bantic: nvm, it is fine

app/password/new/template.hbs
```diff
@@ -0,0 +1,58 @@
...
+            {{#if hasSubmitted}}
+              {{#if errors.model.password}}
+                {{#bs-alert alert="warning"}}
+                  Password {{join-array errors.model.password ', '}}.
+                {{/bs-alert}}
+              {{/if}}
+              {{#if errors.model.passwordConfirmation}}
+                {{#bs-alert alert="warning"}}
+                  Confirmation {{errors.model.passwordConfirmation}}.
+                {{/bs-alert}}
```
bantic: Checking for explicit error properties seems like it could unintentionally omit other unexpected errors, like a `base` error if the API returned one?

app/utils/jwt.js
```diff
@@ -7,7 +7,7 @@ export default Ember.Object.extend({
   payload: function(){
     var token = this.get('token');
 
-    if (!token) { return {}; }
+    if (!token || token === "null") { return {}; }
```
bantic: how does the token end up as `'null`'?

app/validators/password-complexity.js
```diff
@@ -0,0 +1,25 @@
+import Base from 'ember-validations/validators/base';
+
+const COMPLEXITY_RULES = [
+  ['must be at least 10 characters', /^.{0,9}$/],
+  ['must no greater than 128 characters', /^.{129,}$/],
+  ['must contain at least one uppercase letter', /^[^A-Z]+$/],
+  ['must contain at least one lowercase letter', /^[^a-z]+$/],
+  ['must contain at least one digit or special character', /^[^0-9!@#$%^&*()]+$/]
+];
```
bantic: cool.

app/password/new/template.hbs
mixonic: These are validation, not API errors. They don't make a request, I don't believe there is any base.

Can definitely use some DRY, though I'd like to do it separately.

app/utils/jwt.js
mixonic: I'm not sure- bad data gets written at some point. For all I know it was a test or a half-complete implementation of something reloading with live-reload. IMO it seems good to catch when possible.

app/password/new/template.hbs
bantic: Gotcha, thanks. I transposed `errors` and `model`

app/utils/jwt.js
bantic: yeah, just curious. A null is getting stringified somewhere...

bantic:
awesome

app/utils/jwt.js
mixonic: Specifically this is a null value being written to localStorage. That is what stringifies.

PR #130, opened 2015-02-25
UI changes for unverified users

### LINKED ISSUE #128: Unverified account view (screen 201)
from #51 
  
  * [x] Don't block entire UX for unverified users.  Provide glimpse of provisioned account and any "apps" they provisioned in set up flow.
  * ~~[ ] Create the apps/databases but do not provision. Upon verification of a user, provision any apps/dbs they have not provisioned.~~ (already happens as of https://github.com/aptible/diesel.aptible.com/commit/2d1f2aefeaa7d512b379eb4a6f9f4ce0dbb678d6)
  * [x] verify that deploys check for a verified user.
     * Unverified users are prevented from creating operations in https://github.com/aptible/api.aptible.com/pull/168
  * [x] unverified users see a verify page (needs design for apps/dbs they have created)
  * [x] unverified users see a verify message with button to click to resend verification email

bantic:
Unverified users can still use the UI but they see a persistent
'you are not verified, click here to resend verification email' banner.
In addition, the 'create' buttons for apps, dbs and log drains are not show,
and the '/new' pages for each of those three shows a message saying you must be
verified to create.

fixes #128

@mixonic, for you
@sandersonet please take a look at the UI updates (comments on #128) and
let me know if you'd like to see any changes.

### DIFF
--- a/app/apps/new/template.hbs
+++ b/app/apps/new/template.hbs
@@ -1,39 +1,56 @@
-<div class="row">
-  <div class="col-xs-6">
-    <div class="panel panel-default panel-subdued">
-      <div class="panel-heading">
-        <h3>Create a new app</h3>
-      </div>
-      <div class="panel-body">
-        <form role="form">
-          {{#unless model.isValid}}
-            <div class="alert alert-warning fade in">
-              <a class="close" data-dismiss="alert">×</a>
-              {{#each model.errors.messages as |message|}}
-                <p>{{message}}</p>
-              {{/each}}
+{{#if session.currentUser.verified}}
+  <div class="row">
+    <div class="col-xs-6">
+      <div class="panel panel-default panel-subdued">
+        <div class="panel-heading">
+          <h3>Create a new app</h3>
+        </div>
+        <div class="panel-body">
+          <form role="form">
+            {{#unless model.isValid}}
+              <div class="alert alert-warning fade in">
+                <a class="close" data-dismiss="alert">×</a>
+                {{#each model.errors.messages as |message|}}
+                  <p>{{message}}</p>
+                {{/each}}
+              </div>
+            {{/unless}}
+
+            <div class="form-group-last">
+              <label class="block">
+                App handle
+                <span class="label-helper">Lowercase alphanumerics, periods, and dashes only</span>
+              </label>
+              {{handle-input class="form-control" value=model.handle name='handle'}}
             </div>
-          {{/unless}}
+          </form>
 
-          <div class="form-group-last">
-            <label class="block">
-              App handle
-              <span class="label-helper">Lowercase alphanumerics, periods, and dashes only</span>
-            </label>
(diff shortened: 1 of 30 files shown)


app/apps/new/template.hbs
```diff
@@ -1,39 +1,56 @@
...
-        <form role="form">
-          {{#unless model.isValid}}
-            <div class="alert alert-warning fade in">
-              <a class="close" data-dismiss="alert">×</a>
-              {{#each model.errors.messages as |message|}}
-                <p>{{message}}</p>
-              {{/each}}
-            </div>
-          {{/unless}}
+{{#if session.currentUser.verified}}
```
mixonic: I wonder if we can easily refactor this into `renderTemplate` calls on the route.

bantic:
@mixonic Some churn in the test files due to me adding some new test helpers and updating old syntax when I revisited the files.

Lots of churn to the payment test:
  * added a function to mock successful payment and pulled out explicit `fillIn` and assertions for stripe/CC data from tests that were not testing that part
  * changed the user data to default to unverified
  * updated the tests that included provisioning to set the user as verified

app/components/verification-code-reset/template.hbs
```diff
@@ -0,0 +1,27 @@
...
+  <div class="row">
+    {{#bs-alert alert="warning" as |component|}}
+      {{#bs-alert-dismiss target=component}}
+         <button type="button" class="close" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+      {{/bs-alert-dismiss}}
+
+      <span class='danger'>Your email is not activated.</span>
+
+      You will not be able to deploy or create new apps until you activate your email.
+      <button class='btn btn-default btn-xs' {{action 'resendVerification'}} {{bind-attr disabled="isDisabled"}}>
```
mixonic: I think you mean `{{bind-attr disabled=isDisabled}}`, no quotes?

app/templates/-nav-header.hbs
```diff
@@ -10,3 +10,7 @@
     </ul>
   </div>
 </header>
+
+{{#unless session.currentUser.verified}}
+  {{verification-code-reset store=store}}
```
mixonic: Maybe we should just inject store onto components?

app/components/verification-code-reset/template.hbs
bantic: correct, thanks

app/templates/-nav-header.hbs
bantic: done

app/apps/new/template.hbs
bantic: good call, that works well

bantic:
@mixonic addressed your comments in the last two commits. When the tests goes green this is good to go.

bantic:
closes #51

PR #137, opened 2015-02-25
Use deprovision operation to remove VHost

### LINKED ISSUE #123: App#Domain#Delete is not implemented
The delete button on the app vhost index page is not implemented.
It should be implemented to delete the vhost, or removed.

http://take.ms/mjhzO

bantic:
fixes #123

@mixonic, for you

### DIFF
--- a/app/app/vhosts/index/route.js
+++ b/app/app/vhosts/index/route.js
@@ -9,5 +9,25 @@ export default Ember.Route.extend({
     if(model.get('length') === 0) {
       return this.transitionTo('app.vhosts.new', this.modelFor('app'));
     }
+  },
+
+  actions: {
+    delete: function(vhost){
+      let op = this.store.createRecord('operation', {
+        type: 'deprovision',
+        vhost: vhost
+      });
+
+      this.controller.set('error', null);
+      vhost.set('isDeleting', true);
+
+      op.save().then( () => {
+        vhost.deleteRecord();
+      }).catch( () => {
+        this.controller.set('error', 'There was an error deleting the VHost');
+      }).finally( () => {
+        vhost.set('isDeleting', false);
+      });
+    }
   }
 });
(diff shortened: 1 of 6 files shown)


app/app/vhosts/index/route.js
```diff
@@ -9,5 +9,25 @@ export default Ember.Route.extend({
...
+
+  actions: {
+    delete: function(vhost){
+      let op = this.store.createRecord('operation', {
+        type: 'deprovision',
+        vhost: vhost
+      });
+
+      this.controller.set('error', null);
+      vhost.set('isDeleting', true);
```
mixonic: I think this can safely sit on the controller. Hm.

bantic: It definitely makes some sense to make a 'delete-vhost' component that can encapsulate some or all of this logic. At minimum it can encapsulate the button disabling and button text changing during save, removing the necessity of this transient `isDeleting` prop on the vhost.

However because there didn't seem to be a good place to stash an error message (if there is one) on/around that button I added an alert at the top of this `vhosts#index` page, which is what the controller's `error` prop is used for.

Since there are multiple semi-disconnected parts of the UI that are in play here it feels most natural to me to do it this way, even with the ugliness of the `isDeleted` prop on the vhost. The alternative seems like it would be having to coordinate actions on the route and the component to synchronize the states between the two.

I'll give it a shot to see how it feels and update this PR with a delete-vhost component.

bantic: as of 027b883ca11123852e9a22f66a20208ac68a591c we have a store on the components, also

bantic:
@mixonic refactored to use a `{{delete-vhost}}` component in 1b8babcaf881a3bb3afcf3362ba6f124c68c93bf

When tests go green this is good to go

mixonic:
also travis being a miscreant.

bantic:
yeah travis status was reporting some issue w/ public repos

PR #140, opened 2015-02-25
When creating a new log drain, must also create a 'configure' operation

### LINKED ISSUE #138: After creating a log drain, create a 'configure' operation
Refs #129 

The correct flow for diesel to take is:

  * post to the API to create a log drain
  * post to the API to create a 'configure' operation with that log drain as its resource

bantic:
fixes #138

@mixonic for you


mixonic:
Travis had a hiccup here and is not reporting that this passes. But it does.

PR #155, opened 2015-03-16
Upgrade to ember 1.11.0-beta.5

bantic:
upgrade to [ember 1.11.0-beta.5](http://emberjs.com/blog/2015/02/07/ember-1-10-0-released.html)

### DIFF
--- a/app/components/aptible-ability/component.js
+++ b/app/components/aptible-ability/component.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  hasAbility: false, // cannot use `isVisible` with `tagName: ''`
+  hasAbility: false,
   tagName: '', // tagName must be exactly '' to render this component with no surrounding tag
 
   user: null,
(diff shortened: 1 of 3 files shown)


bantic:
I'm going to merge this after the build goes green.

mixonic:
@bantic `tagName: ''` didn't work?

bantic:
@mixonic It breaks the test with "Assertion Failed: You cannot access this.$() on a component with `tagName: ''` specified."

mixonic:
That is unexpected. I wonder if that means an addon is reopening all components or something. We should still be able to use `tagName: ''`

bantic:
@mixonic it comes from here: https://github.com/emberjs/ember.js/blob/06e41ad7ccd28558dd4e651aa070bc06f7757821/packages/ember-views/tests/views/view/jquery_test.js#L54

Ok to merge, or do you want to look into that?

bantic:
Pending this upstream PR to change `moduleForModel`'s `render` callback:
https://github.com/switchfly/ember-test-helpers/pull/26

mixonic:
@sandersonet we believe much of the back button funkiness was due to a bug in Ember 1.10. It is fixed in 1.11, and we will check how things improve with this patch.

PR #156, opened 2015-03-16
Signup validations

sandersonet:
This PR adds a component for contextual input validation feedback: 

screenshot 2015-03-16 11 55 21

A follow up to this PR would be to add remote validators for user#email, app#handle, account#handle, and database#handle

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -7,8 +7,8 @@ module.exports = function(environment) {
     baseURL: '/',
     locationType: 'auto',
 
-    authBaseUri: "http://localhost:4000",
-    apiBaseUri: "http://localhost:4001",
+    authBaseUri: "http://auth.aptible.foundry.io",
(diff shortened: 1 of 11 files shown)


config/environment.js
```diff
@@ -7,8 +7,8 @@ module.exports = function(environment) {
     baseURL: '/',
     locationType: 'auto',
 
-    authBaseUri: "http://localhost:4000",
-    apiBaseUri: "http://localhost:4001",
```
mixonic: Would like to leave these as the localhost variants, unless there is a reason to change them.

sandersonet: Oops, good catch.

mixonic:
Don't want to hold up on this, but we should also squash down commits like "fix tests". Make the history easier to review and read when things go wrong.

PR #166, opened 2015-03-17
Focus first form input for apps/new, dbs/new, vhosts/new, logs/new

### LINKED ISSUE #146: Focus on first input when entering a resource/new route

bantic:
* add autofocusable-select view
  * add Autofocusable mixin to reclaim focus when input is inserted
  * mix Autofocusable into autofocusable-select and handle-input
  * add `expectFocusedInput` test helper

fixes #146 

@mixonic  for you, cc @sandersonet 

Learned some interesting things about the html `autofocus` boolean attr while doing this. Namely that it doesn't work as well as it should, esp on FF and Chrome. When an input is added to the DOM via JavaScript it doesn't always get focused even with the autofocus attribute set. That's the reason for the `Autofocusable` mixin.

Example [jsbin demonstrating this with ember](http://jsbin.com/tasiqo) and [jsbin demonstrating without ember](http://jsbin.com/hewage#). Doing it this way is somewhat hacky but I think a good balance between a more explicit solution (e.g., a view specifically for each route that finds and focuses the input) and more hacking to fix the buggy browser autofocus. If this were a mission-critical feature rather than an enhancement I would suggest a bit more work making it bulletproof.

cc @sandersonet

### DIFF
--- a/app/app/vhosts/new/template.hbs
+++ b/app/app/vhosts/new/template.hbs
@@ -17,10 +17,11 @@
 
           <div class="form-group">
             <label>Service</label>
-            {{view "select" content=services
+            {{view "autofocusable-select" content=services
                             selection=vhostService
                             name="service"
                             optionLabelPath="content.handle"
+                            autofocus=true
                             class="form-control"}}
           </div>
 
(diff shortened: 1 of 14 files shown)


bantic:
This PR also puts us over 1000 tests. :fireworks:

mixonic:
@bantic needs rebasing on top of your other PR, but very good thanks!

bantic:
rebasing now

bantic:
when the tests go green I will merge this in

PR #171, opened 2015-03-18
Disable login button while user is logging in

bantic:
This makes the login button disabled (and its message changes to "Logging you in...")  while the user is being logged in.
A UI enhancement and prevents this error that can occur if torii attempts to log a user in while they are already logging in (double-clicking the login button will cause this, for example)

login

(edited to fix gif above)

cc @sandersonet 
I'll merge this in when the tests pass.


mixonic:
Awesome.

Issue #170, opened 2015-03-18
Password reset error should blank when you leave the route


mixonic:
Currently the error sticks around forever. Controllers are singletons!

PR #177, opened 2015-03-18
ensure that error messages for email/password change are cleared

### LINKED ISSUE #176: password/email change error should blank when leaving route
similar to #170, when an error is shown on /settings/profile or /settingsadmin/, that error should disappear when re-entering the route.

bantic:
* cleans up some tests to use more es6 and aptibles `expectInput`, `expectButton` etc helpers
  * tests for email/password change clearing errors on accounts/admin and accounts/profile
  * some modification to the tests for password/reset clearing errors

refs #170
fixes #176

### DIFF
--- a/app/password/reset/route.js
+++ b/app/password/reset/route.js
@@ -2,18 +2,24 @@ import Ember from 'ember';
 import DisallowAuthenticated from "../../mixins/routes/disallow-authenticated";
 
 export default Ember.Route.extend(DisallowAuthenticated, {
-  model: function(){
+  model() {
     return this.store.createRecord('password-reset-request');
   },
 
+  resetController(controller) {
+    controller.setProperties({
+      error: null,
+      hasSubmitted: false
+    });
+  },
+
   actions: {
-    reset: function(model){
+    reset(model) {
       model.save().then( () => {
         this.transitionTo('login');
       }, () => {
-        this.controllerFor('password/reset').set('error', `
-          There was an error resetting your password.
-        `);
+        this.controller.set(
+          'error', `There was an error resetting your password. `);
       });
     },
     willTransition() {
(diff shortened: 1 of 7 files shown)


bantic:
cc @mixonic lmk if you want to review. I'll merge in this afternoon if not, since these are fairly minor changes. Mostly es6-ification and test cleanup

app/password/reset/route.js
```diff
@@ -2,18 +2,26 @@ import Ember from 'ember';
 import DisallowAuthenticated from "../../mixins/routes/disallow-authenticated";
 
 export default Ember.Route.extend(DisallowAuthenticated, {
-  model: function(){
+  model() {
     return this.store.createRecord('password-reset-request');
   },
 
+  resetController(controller, isExiting) {
```
mixonic: Whoa- I used `willTransition`. This is better, I didn't realized it had been added.

mixonic: In fact I don't think you need to check the `isExiting` boolean, right? b/c you would want these things to be reset even if there was a transition to this same route.

tests/acceptance/settings/settings-account-test.js
```diff
@@ -225,18 +214,28 @@ test('visit ' + settingsAccountUrl + ' change email and errors', function(){
...
-    var error = find('.alert');
+    let error = find('.alert');
     ok(error.length, 'shows error div');
     ok(error.text().indexOf('Invalid password'), 'shows error message');
   });
+
+  andThen( () => {
+    visit(settingsProfileUrl); // go away
+    visit(settingsAccountUrl); // come back
+  });
```
mixonic: These don't need to be inside an `andThen`

bantic: Yeah, I think this improves test legibility because it groups the leave-and-return step from the error-checking assertions above it. If that seems unneeded these can go in the above andThen

app/password/reset/route.js
bantic: yeah, good point, I'll fix that

PR #178, opened 2015-03-18
Various style improvements

sandersonet:
* Shortens the viewable git ref to first 10 characters.  The entire ref is viewable in a tooltip and copyable with click-to-copy component
* Added click-to-copy to other long fields: git ref, git remote, and DB connection URL
* Increased font size for resource meta
* Hides app nav if the app has never been deployed

cc @mixonic

### DIFF
--- a/app/components/bs-tooltip/component.js
+++ b/app/components/bs-tooltip/component.js
@@ -22,6 +22,8 @@ export default Ember.Component.extend(BootstrapComponentOptions, {
     var options = this.getBootstrapOptions();
     var tooltip = this.$(':first-child').tooltip(options);
 
-    if(options.trigger = 'immediate') { tooltip.tooltip('show'); }
+    if(options.trigger === 'immediate') {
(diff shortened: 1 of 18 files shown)


app/components/bs-tooltip/component.js
```diff
@@ -22,6 +22,6 @@ export default Ember.Component.extend(BootstrapComponentOptions, {
     var options = this.getBootstrapOptions();
     var tooltip = this.$(':first-child').tooltip(options);
 
-    if(options.trigger = 'immediate') { tooltip.tooltip('show'); }
+    if(options.trigger === 'immediate') { tooltip.tooltip('show'); }
```
mixonic: Just a style nit, but using the multi-line version of `if` is preferred, since you can then more easily add a debugger to a line. For example, adding a debugger to the `tooltip` function here would be tricky.

sandersonet: Thanks!

PR #182, opened 2015-03-19
Use an in-memory hash as a shim for localStorage when it isn't supported

### LINKED ISSUE #169: localStorage fails in Safari private mode
When trying to log in from a Safari Private Browsing window, you'll see this error:

image

Safari appears to set the localstorage quota to 0 in private browsing mode. We should fallback to an cookies in this case, most likely. Or in-memory store.

bantic:
Safari in private browsing mode sets a quota of 0 so we cannot
use localStorage.setItem in private safari.

fixes #169

@mixonic for your review

### DIFF
--- a/app/utils/storage.js
+++ b/app/utils/storage.js
@@ -1,5 +1,35 @@
 var serializedPrefix = '_object:';
 
+let NativeAdapter = function(){
+  this.getItem = (key) => window.localStorage.getItem(key);
+  this.setItem = (key, val) => window.localStorage.setItem(key, val);
+  this.removeItem = (val) => window.localStorage.removeItem(val);
+};
+
+let MemoryAdapter = function(){
+  let storage = {};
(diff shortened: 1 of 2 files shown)


app/utils/storage.js
```diff
@@ -1,5 +1,35 @@
...
+
+let MemoryAdapter = function(){
+  let storage = {};
+  this.getItem = (key) => storage[key];
+  this.setItem = (key, val) => storage[key] = val;
+  this.removeItem = (val) => delete storage[val];
+};
+
+function checkLocalStorage(){
+  try {
```
mixonic: Checking for:

```
if (!window.localStorage) {
  return false;
}
```

also seems reasonable for Very Old browsers (pre IE10)

mixonic: It will caught anyway, just avoids a try/catch if unneeded.

PR #189, opened 2015-03-23
Updated layout for stack, app, database, and settings

sandersonet:
cc @mixonic @bantic 

This PR addresses some of the design feedback from last week in a couple of ways:
* Add a sidebar to stack.show, providing linkage to other stacks and organization settings
* Add resource sidebar to resource.show.* routes, allowing cross linking to other resources in stack and breadcrumb to parent stack
* Adds resource icon to header of each resource type
* Stronger app layout with flexbox-based holy grail layout

#### App Show:
screenshot 2015-03-23 10 29 44

#### Database Show:
screen shot 2015-03-23 at 10 38 31 am

#### Stack Show: 
screenshot 2015-03-23 10 29 22

#### Verification Alert:
screenshot 2015-03-23 10 29 05


app/components/resource-sidebar/template.hbs
```diff
@@ -0,0 +1,48 @@
...
+      {{#if hasAbility}}
+        {{#link-to 'apps.new' stack }}
+          <i class="fa fa-plus"></i>
+        {{/link-to}}
+      {{/if}}
+    {{/aptible-ability}}
+  </li>
+
+  {{#each stack.apps as |app|}}
+    {{#link-to 'app' app tagName="li"}}
```
bantic: We added an `activating-item` component in  9aa5a2a15d756e39bddc1f1b31f1e01a19f5ba8e, you could use it here as:
```
{{activating-item currentWhen="app" tagName="li"}}
  {{link-to app.handle 'app' app}}
{{/activating-item}}
```

app/components/resource-sidebar/template.hbs
```diff
@@ -0,0 +1,48 @@
...
+      {{link-to app.handle 'app' app bubbles=false}}
+    {{/link-to}}
+  {{/each}}
+</ul>
+
+<ul class="nav sidebar-nav">
+  <li class="title">
+    Databases
+
+    {{#aptible-ability scope="manage" user=user stack=stack as |hasAbility|}}
```
bantic: `aptible-ability` changed in 95c02b60e290cda4be36f66e43def160ae7341eb to have a `permittable` param instead of stack, so this should be:
`{{#aptible-ability scope="manage" user=user permittable=stack as |hasAbility|}}`

app/components/resource-sidebar/template.hbs
mixonic: In fact, need to use it to work around a bug in 1.11-beta.5

app/stack/route.js
```diff
@@ -14,17 +14,17 @@ export default Ember.Route.extend({
     return tokens.join(' - ');
   },
 
+  setupController: function(controller, model) {
+    controller.set('model', model);
+    controller.set('organizations', this.store.find('organization'));
```
bantic: since `setupController` is a synchronous callback (as opposed to async callbacks like `model`, `afterModel`, `beforeModel`), but `this.store.find('organization')` is an async call, the controller will has `organizations` set to a promise proxy wrapping the in-flight ajax call for organizations. As soon as the API responds the content will be made available to the template but until then the stack template will have an empty space where the organizations would be. And since the stack sidebar relies on the stacks, it may also be empty for a moment for each organization if the `store.find('organizations')` request comes back significantly before the `store.find('stacks')` request does.

This is probably fine as it is, just wanted to clarify that that is what is going on here. It most likely makes the most sense to let this template start rendering before the orgs and stacks have been loaded.

If you wanted to get fancy you could check the `isLoading` property of `organizations` and I believe something like this would work (in the stack/template.hbs):
```
{{#each organizations as |organization|}}
  stack-sidebar etc
{{else if organizations.isLoading}}     {{! the store.find ajax call is in-flight}}
  loading orgs...
{{else}}     {{! not loading anymore and the organizations.length is 0}}
  you have no organizations
{{/each}}
```

bantic:
cool to see flexbox in action!

app/components/resource-sidebar/template.hbs
sandersonet: I will leave this for now because it "works", but we'll need to update it to use `activating-item` once https://github.com/aptible/diesel.aptible.com/issues/196 is implemented.

app/stack/route.js
sandersonet: @bantic Thanks for the explanation.  Since they are only used for the sidebar, I think in this instance it's good to have organizations and stacks loaded async so as to not block the rendering of the primary content.  Please let me know if there is a better way to handle this.

app/app/route.js
```diff
@@ -15,8 +15,12 @@ export default Ember.Route.extend({
     return tokens.join(' - ');
   },
 
+  setupController: function(controller, model){
+    controller.set('model', model);
+    controller.set('stack', model.get('stack.content'));
```
sandersonet: > model.get('stack.content`)

@bantic  Getting stack.content still seems to be required here in order to make sure `permittable` works.  Am I missing a better way to handle this?

bantic: @sandersonet Yeah, sorry I didn't describe that better before. Because app.stack is an async relationship (the ember-data relationship is [defined here](https://github.com/aptible/diesel.aptible.com/blob/master/app/models/app.js#L19)), app.get('stack') is *always* going to return a promise proxy (whether the stack is loaded or not). This usually works fine in templates because the it will proxy all `get`s for properties to the content. But aptible-ability is checking for a `permitsRole` method directly on the promise proxy, which isn't there.

Usually it's an anti-pattern to access the content directly, because if the async relationship isn't loaded `app.get('stack.content')` will equal `undefined`.

Because of the `afterModel` call in this route, we can be certain that the app's stack is loaded, so `app.get('stack.content')` returns the actual instance of the `Stack` model rather than undefined.

There are 2 ways to get around accessing content directly that I can think of:
  * make aptible-ability "promise-aware", having it absorb that async and defer checking the ability until everything passed to it is loaded. This is a little complicated but if it makes ability-checking smoother it may be worthwhile.
  * stash the stack as a property on the route in afterModel:
    ```
      afterModel(model){
         return model.get('stack').then( (stack) => {
            this.stack = stack;
          });
          // in setupController, you can now do "controller.set('stack', this.stack);"
      }
   ```
   This isn't ideal because you are adding statefulness to the router.

cc @mixonic Any other ideas for best practices here?

app/stack/route.js
bantic: That makes sense to me, too. My comment was there to make it clear that the tradeoffs/differences between the approaches so you can choose the one that fits the use case best.

app/app/route.js
mixonic: IMO making `aptible-ability` async-aware would be the ideal. That sounds like a good direction.

But I think this is good to go if you simply add a `FIXME` on the previous line explaining that the `stack` is loaded by the `afterModel` hook, and that accessing `content` looks into the promise proxy for the real relationship model.

sandersonet:
@bantic @mixonic I think this is ready, I appreciate your feedback.

app/components/stack-sidebar/component.js
```diff
@@ -3,5 +3,5 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   organizationStacks: function() {
     return this.get('stacks').filterBy('organization.id', this.get('organization.id'));
-  }.property('stacks.@each.organization.id', 'organization.id')
-});
+  }.property('stacks.[]', 'stacks.@each.organization.id', 'organization.id')
```
mixonic: `stacks.[]` is not needed if `stacks.@each.organization.id` is present.

app/components/verification-code-reset/template.hbs
```diff
@@ -1,29 +1,23 @@
...
-            Sending
-          {{else if sent}}
-            Verification email sent
-          {{else}}
-            <i class='fa fa-refresh'></i>
-            Resend verification email
-          {{/if}}
-        </button>
+  You will not be able to deploy or create new apps until you activate your email.
+  <button class='btn btn-default btn-xs' {{action 'resendVerification'}} {{bind-attr disabled=isDisabled}}>
```
mixonic: `disabled={{isDisabled}}` :-D

mixonic:
I've made the two changes I called out and merged this into master manually. Closing this PR.

PR #192, opened 2015-03-23
Allow database type to be unselected during welcome

### LINKED ISSUE #186: No way to unselect database type in welcome/first-app
Since database is optional, we should allow a user to unselect the current selection.  The simplest way to do this would be to just have a click event on the selected DB type that unselects the type and removes the database model/UI.   

We should only apply this fix to `welcome.first-app`.  Since database type is required in `database.new`, we should ensure that a type is alway selected (how it currently works).

mixonic:
Fixes #186.

Data-down, actions up! It works! :-D

### DIFF
--- a/app/welcome/route.js
+++ b/app/welcome/route.js
@@ -1,5 +1,11 @@
 import Ember from 'ember';
 
+export function resetDBData(model){
+  Ember.set(model, 'dbType', null);
+  Ember.set(model, 'initialDiskSize', 10);
+  Ember.set(model, 'dbHandle', '');
+}
+
 export default Ember.Route.extend({
   beforeModel: function(){
     if(this.session.get('isAuthenticated')) {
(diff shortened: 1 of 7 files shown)


mixonic:
@sandersonet for you to review/merge please!

app/welcome/route.js
```diff
@@ -1,5 +1,11 @@
 import Ember from 'ember';
 
+export function resetDBData(model){
+  Ember.set(model, 'dbType', null);
+  Ember.set(model, 'initialDiskSize', 10);
```
sandersonet: @mixonic What is the advantage of using `Ember.set` versus just `model.set` ?  I'm also curious if `resetDBData` would be better as an instance method on the database model?

mixonic: `Ember.set` can be called an Ember object, created with `Ember.Object.create`:

```
var obj = Ember.Object.create();
Ember.set(obj, 'prop', val);
```

Or with a POJO:

```
var obj = {};
Ember.set(obj, 'prop', val);
```

The POJO would not have `obj.set`, since it is just a POJO. The `set` is required since we are observing these properties elsewhere, like in the templates. This is a pretty bare data payload or configuration object, a data-bag, and I didn't feel it was appropriate to use an Ember Object (which is preferred for class/instance usage).

The POJO model for `welcome` is only used in the routing. It is not a `database` model instance.

sandersonet: Awesome, that makes sense.  Thanks for the explanation.

PR #203, opened 2015-03-24
Use feature flag for cost estimator visibility fixes #202

### LINKED ISSUE #202: Hide pricing estimator
Probably behind a feature flag

sandersonet:
cc @mixonic

### DIFF
--- a/tests/unit/components/estimated-cost-test.js
+++ b/tests/unit/components/estimated-cost-test.js
@@ -3,41 +3,61 @@ import {
   test
 } from 'ember-qunit';
 
+function mockFeaturesService(isEnabled) {
+  return {
+    isEnabled: function(feature) {
+      return isEnabled && feature === 'price-estimator';
+    }
+  };
+}
+
 import Ember from 'ember';
 
-moduleForComponent('estimated-cost', 'EstimatedCostComponent', {});
+moduleForComponent('estimated-cost', 'EstimatedCostComponent', {
+  subject: function(opts = {}) {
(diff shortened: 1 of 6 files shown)


tests/unit/components/estimated-cost-test.js
```diff
@@ -3,6 +3,18 @@ import {
   test
 } from 'ember-qunit';
 
+var stubFeature = function(component, feature, value) {
+  component.features = {
```
sandersonet: @mixonic Seems like feature stubbing is built into acceptance tests, but not unit tests.  Is there a better way to do this?

mixonic: `feature` is normally present via dependency injection. One of the reasons we use DI is because it and easily be implemented with composition. Feature is a good example.

Since there are no injections in unit tests, use `create` to pass the service (this is composition):

```javascript
function mockFeaturesService(){
  return {
    isEnabled: function(feature) {
      return feature === 'price-estimator';
    }
  };
}

test('something something', function() {
  let component = this.subject({
    features: mockFeaturesService(),
    count: 5
  })
});
```

I'd suggest doing this over modifying the component after the fact, since you want the property to be there when `init` is called on the component.

If the same service is injected for every test, it can be a little more DRY to use the `subject` hook during setup. For example [this code](https://github.com/aptible/diesel.aptible.com/blob/486dfeb083dc9cd0dd128a7f750c8102f1f6043e/tests/unit/components/query-box/component-test.js#L13-L18) adds a mock analytics service to a component.

PR #211, opened 2015-03-24
Add a confirmation modal and use it for deleting roles

### LINKED ISSUE #206: Organization > Delete Role

mixonic:
Add a service that can be used for triggering a confirmation modal. For example:

```javascript
  confirmationModalService = Ember.inject.service('confirmation-modal'),
  actions: {
    delete(thing) {
      this.get('confirmationModalService').open({
        partial: 'confirmation-modals/some-partial-template', // Is looked up as a partial
        model: thing, // Is made available in the partial
        onConfirm() {
          thing.deleteRecord();
          return thing.save(); // The modal will not close until a returned promise has completed
        } // onCancel also available
      });
    }
  }
});
```

Uses this for role deletion.

Fixes #206

### DIFF
--- a/app/components/confirmation-modal/component.js
+++ b/app/components/confirmation-modal/component.js
@@ -1,0 +1,44 @@
+import Ember from 'ember';
+import layout from './template';
+
+export default Ember.Component.extend({
+  confirmationModalService: Ember.inject.service('confirmation-modal'),
+  layout: layout,
+
+  model: Ember.computed.reads('modal.model'),
+
+  startConfirmationModalServiceListener: function() {
+    this._super.apply(this, arguments);
+    this.get('confirmationModalService').on('open', (modal) => {
+      this.set('modal', modal);
+      this.set('isOpen', true);
+    });
+  }.on('init'),
+
+  close() {
+    this.set('isOpen', false);
+    this.set('modal', null);
+  },
+
+  actions: {
+
+    cancel() {
+      let modal = this.get('modal');
+      new Ember.RSVP.Promise((resolve) => {
+        resolve(modal && modal.onCancel && modal.onCancel());
+      }).then(() => {
+        this.close();
+      });
+    },
+
+    confirm() {
+      let modal = this.get('modal');
+      new Ember.RSVP.Promise((resolve) => {
+        resolve(modal && modal.onConfirm && modal.onConfirm());
+      }).then(() => {
+        this.close();
+      });
+    }
+
+  }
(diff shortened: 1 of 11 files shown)


mixonic:
@bantic for you to review/merge please.

sandersonet:
Really cool @mixonic

app/components/confirmation-modal/component.js
```diff
@@ -0,0 +1,46 @@
...
+
+  model: Ember.computed.reads('modal.model'),
+
+  init() {
+    this._super.apply(this, arguments);
+    this.get('confirmationModalService').on('open', (modal) => {
+      this.set('modal', modal);
+      this.set('isOpen', true);
+    });
+  },
```
bantic: small nitpick: would be less potential footgun to use the `on('init')` hook instead of overriding init. Also provides an opportunity to use an intention-revealing function name like `setupConfirmationModal`

app/components/confirmation-modal/component.js
```diff
@@ -0,0 +1,46 @@
...
+  actions: {
+
+    cancel() {
+      let modal = this.get('modal');
+      let result = modal.onCancel && modal.onCancel();
+      if (result && result.then) {
+        result.then(() => this.close());
+      } else {
+        this.close();
+      }
```
bantic: Would it be simple to promisify the result instead of detecting that it is a thennable?
```
return Ember.RSVP.resolve().then(() => {
  return result;
}).then(() => {
  this.close();
});
```

app/organization/roles/template.hbs
```diff
@@ -43,7 +43,7 @@
                     </li>
                     <li>
                       {{! FIXME should fire action that calls deletion }}
```
bantic: this FIXME can go now

bantic:
Cool, a few notes. Neat idea. @mixonic @sandersonet Do you think users will find it jarring to have the modal replace the entire UI, rather than laying on top of it?

app/components/confirmation-modal/component.js
mixonic: So, the future of `on('init')` type callbacks seems unclear. I don't think there is much interest in removing them.

There might be some curse of expertise going on here, I definitely have a gut feeling that I want to use `init() {` instead of an evented hook. Not sure why.

mixonic: Unsure. Changing to the hook for now though.

PR #215, opened 2015-03-24
Allow sending an invitation reset

### LINKED ISSUE #204: organizations > resend invitation
refs #188 

There is a "refresh" icon next to the list of invited members. Clicking it should fire an invitation reset to resend the invitation to that user.

bantic:
* show success message on completion
  * Add resend-invitation component

fixes #204

### DIFF
--- a/app/components/resend-invitation/component.js
+++ b/app/components/resend-invitation/component.js
@@ -1,0 +1,24 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  tagName: '', // tagless component
+
+  store: Ember.inject.service(),
+  invitation: null,
+
+  actions: {
+    resend(){
+      let invitation = this.get('invitation');
+      let invitationId = this.get('invitation.id');
+      let reset = this.get('store').createRecord('reset');
+      reset.setProperties({
+        type: 'invitation',
+        invitationId
+      });
+
+      reset.save().then( () => {
+        this.sendAction('action', invitation);
+      });
+    }
+  }
+});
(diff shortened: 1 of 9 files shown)


app/components/resend-invitation/component.js
```diff
@@ -0,0 +1,24 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  tagName: '', // tagless component
+
+  store: Ember.inject.service(),
```
mixonic: components already have the store via:

https://github.com/aptible/diesel.aptible.com/blob/master/app/initializers/setup-component-with-store.js#L2

this injection would fail if you used `this.get('store')` as the store is not a service (not in the container as such). The lazy injections with `Ember.inject` need to be fetched via `.get`, and since you used `this.store` below this injection is actually never used.

bantic: ah, good point. It probably makes sense to remove the global setup-component-with-store initializer in favor of targeted injections a la `Ember.service.inject`, though, yeah?

bantic: The store is a service as of ember data 16, which we are using: http://emberjs.com/blog/2015/03/23/ember-data-1-0-beta-16-released.html#toc_store-as-a-service

mixonic: Ace. I figured there might be something like that. Dem release notes!

I think targeted injections make more sense yeah. But you will need to use `get` again.

bantic: I'm going to fix this and will make a separate pr to remove the global initializer

PR #216, opened 2015-03-24
Allow destroying an invitation

bantic:
For after #215

Wires up the "X" button to destroy an invitation

### DIFF
--- a/app/organization/invitations/route.js
+++ b/app/organization/invitations/route.js
@@ -15,6 +15,13 @@ export default Ember.Route.extend({
     resentInvitation(invitation){
       let successMessage = `Re-sent invitation to ${invitation.get('email')}`;
       this.controller.set('successMessage', successMessage);
+    },
+
+    destroyInvitation(invitation){
+      invitation.destroyRecord().then(() => {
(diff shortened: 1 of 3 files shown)


bantic:
@mixonic rebased after #215 and could use a look-over if you would like.

I didn't make a `{{destroy-invitation}}` component here like the `{{resend-invitation}}` component, opting to let the route's `destroyInvitation` action do the async and UI work, because unlike the resend-invitation, destroy-invitation isn't idempotent. By the time the `invitation.destroyRecord()` promise gets to its `then`, ember will have pulled that invitation from the controller's model and this row of the table (and the component) will have been removed. In particular, this makes a `sendAction` from within the component tricky because its `targetObject` has been nullified at that point.

I toyed with closing over the component's target object but it ended up feeling like too much ember inside baseball, and I let the route's action handle everything.

I don't love the side-by-side `resentInvitation` (past tense) and `destroyInvitation` (imperative) actions in the route, though.

mixonic:
dyin' to get new actions. Maybe we can implement them next week, it would need to be on top of Glimmer though.

Issue #213, opened 2015-03-24
Custom 404


sandersonet:
Looks like Divshot is serving up a standard 404.  For now, lets use the existing 404:  https://dashboard.aptible.com/404

bantic:
@sandersonet Do you mean a catchall route for ember to display when it doesn't recognize the route? Right now [divshot is configured](https://github.com/aptible/diesel.aptible.com/blob/master/divshot.production.json#L8) to serve the index.html for any route it doesn't recognize.

bantic:
closed by #218

PR #218, opened 2015-03-24
Add a catchall route and show a 404 message on it

bantic:
refs #213

cc @sandersonet


bantic:
@sandersonet Is this what you had in mind for #213?

Any url that ember attempts and fails to route will display this not-found template.

Looks like this:
image

sandersonet:
:+1: Perfect!

app/templates/shared/-not-found-error.hbs
```diff
@@ -0,0 +1,120 @@
...
+    {{message}} Maybe the URL you visited is no longer valid?
+  {{else}}
+    Maybe you mistyped the address? If not, the page may have moved.
+  {{/if}}
+  </p>
+</div>
+<ul>
+  <li><a href="http://status.aptible.com">Aptible Status</a></li>
+  <li><a href="https://twitter.com/aptiblestatus">@aptiblestatus</a></li>
+  <li>{{#link-to-aptible app="status"}}Contact Support{{/link-to-aptible}}</li>
```
sandersonet: Should this be `app="support"` ?

bantic: thanks, good catch. I fixed this and added an assertion in `link-to-aptible` that should help catch future mistakes with the app name.

PR #222, opened 2015-03-25
Regression fixes

sandersonet:
* Fixes an issue where the confirmation-modal component wraps an element inside the flexbox layout, breaking the ideal max width.
* Fixes an issue in aptible-ability that doesn't check if a role is privileged when determining stack authorization

cc @bantic


bantic:
@sandersonet Great, thanks for catching that. cc @mixonic he and I missed this when modifying aptible-ability to handle organizations

sandersonet:
No problem.  I've updated the specs to check for this use case.

PR #226, opened 2015-03-25
bs-alert sends an action when it is dismissed

### LINKED ISSUE #223: bs-alert needs to reset its visibility state to allow for multiple messages
The `bs-alert` component will flip to a "hidden" state when the user clicks the `bs-alert-dismiss` text. After this the component will stay in a hidden state even if it needs to be displayed again.

For example, on the org settings > invitations page, a user may click the "x" to delete multiple invitations. If the user dismisses the first success message ("deleted invitation to ${email}"), no further success messages will appear because the bs-alert component will stay in hidden state.

bantic:
This allows consumers to react to the dismissal, and helps in cases
where we are reusing the same bs-alert with different messages, to
ensure that the component is torn down and re-rendered, rather than
staying around stuck in its hidden state.

fixes #223

@mixonic for your review, in case there's a better way to "reset" this component.
If it is possible for it to observe its template contents then it could flip
from hidden -> visible when the contents change. I'm not sure if that's easy to do, though.

### DIFF
--- a/app/components/bs-alert-dismiss/component.js
+++ b/app/components/bs-alert-dismiss/component.js
@@ -11,10 +11,10 @@
      {{/bs-alert}}
   {{/bs-alert}}
 
-  * If used in inline style, this will insert a pulled-right "X":
+  * If used in inline style, this will insert a pulled-right "X" (this is the bootstrap default)
 
    {{#bs-alert as |component|}}
-     {{bs-alert-dismiss target=component}}
+     {{bs-alert-dismiss target=component}} {{! inline style}}
    {{/bs-alert}}
  */
 import Ember from 'ember';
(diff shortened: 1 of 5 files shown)


mixonic:
My only thought is that the component could just hide. The "dismissed" state is directly related to the content being displayed- so once I hide a given message the UI can stay hidden until the success message changed.

I'm not sure if clearing the success message on the controller is really necessary. It probably should be cleared on `willTransition`, but actually clearing it when the user clicks dismissed doesn't feel necessary.

bantic:
The component itself isn't torn down, which is part of the problem. There's a little more context in #223.

A better solution (to come) will be to use a flash-style service, like ember-cli-flashes.

Discussed w/ @mixonic in person, merging

Issue #225, opened 2015-03-25
Organizations > edit permission on a role


mixonic:
In the old UI:

![](https://api.monosnap.com/rpc/file/download?id=CygjH8J7TxIb3fhtyCgeFAllyCfIxS)

The list of checkboxes is implemented as follows:

* Walk all accounts/stacks/environments for this organization
* For each, display the name and:
  * A checkbox for a permission named manage
  * A checkbox for a permission name read
* If a box is checked, this means the permission should be created
* If a box in unchecked, that permission should be destroyed.

refs #188

bantic:
this is now part of the roles/show page and is covered by #239

PR #231, opened 2015-03-26
Edit role permissions

mixonic:
Introduce editing permissions attached to a role. This uses a new utility called Changeset.

Changeset is designed to help us manage arbitrary changes to the state of a relationship (or even a property, anything really) that exist only in memory and on the screen and may be disregarded. The
pattern is to only address the relationship via a key made of the related objects, and change a value in memory. When the user is ready to save, this list of changed can be iterated and each staged change can be inspected against its initial value to determine what should be done (ala create a relationship, delete a relationship, etc).

After being configured, a changeset can be [queried for values](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/components/permission-checkbox/component.js#L37), [have new values set](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/components/permission-checkbox/component.js#L44-L46), and [be watched for changes to a value](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/components/permission-checkbox/component.js#L31). This API intentionally mirrors that of Streams in Ember, since the concept is very similar.

Setting up a changeset means [describing how to build a unique key](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/organization/roles/edit/route.js#L33-L35) from the arbitrary key data passed in, and writing an [initialValue](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/organization/roles/edit/route.js#L36-L52) hook that describes how to convert keyData into the initial value.

After making changes, all possibly-modified staged objects (those that have had their values read or set) can be [walked with forEachValue](https://github.com/aptible/diesel.aptible.com/blob/93ba03af38594ef9b83cd003e4c46465692c12f5/app/organization/roles/edit/route.js#L62). This hook is provided the key data, initial value and the final value.

By deferring the walk via forEachValue, we defer making any changes to Ember-Data (or an alternative backend) until the moment of "save". At anytime, we can throw away the whole changeset and nothing outside it has actually been changed.

The [unit tests](https://github.com/mixonic/diesel.aptible.com/blob/afacc4c2ae3f480f8072ad8a6ceb9be910e4e3ef/tests/unit/utils/changeset-test.js) may provide a helpful guide to intended use.


bantic:
this fixes #225

PR #244, opened 2015-03-30
Refactor role edit -> show

mixonic:
@sandersonet I'm going to merge this into master. It changes the roles/edit url to be a roles/show path instead. Seems good to merge before we do any major overhauls on these pages.


sandersonet:
:+1:

PR #255, opened 2015-04-02
Validation feedback on submit

conrad-vanl:
Attempted to sign up, didn't realize "Organization Name" was a required field, clicking submit seemed unresponsive as no feedback was given that there was a validation error. Thought I'd make a quick fix.

Fixed by changing the logic that determines when an error message is shown so that they appear when a form is submitted (and has errors). This works by passing the `hasSubmitted` property to the `validation-feedback` component. 

Note: the name of the attribute on the component is currently `hasSubmitted` to match the `signup` controller. However I am unsure if this name is fitting for the component - perhaps it should be called something like `showError` (or something else...) instead.

Also added `errorText` computed property to `validation-feedback` component to better handle displaying multiple errors at once.


app/components/validation-feedback/component.js
```diff
@@ -2,6 +2,22 @@ import Ember from 'ember';
...
+  /** @type {Boolean} Whether errors will show even if no value */
+  hasSubmitted: false,
+
+  shouldShowFeedback: Ember.computed.or("value", "hasSubmitted"),
+
+  errorText: function() {
+    var error = this.get("error");
+
+    // If multiple errors, join errors with space and comma
+    if(Array.isArray(error) && error.join) {
```
mixonic: Ember has an `isArray` that is ES3-safe fyi ([docs](http://emberjs.com/api/#method_isArray)), but we don't support IE8 anyway :-D

app/components/validation-feedback/component.js
```diff
@@ -2,6 +2,22 @@ import Ember from 'ember';
...
+      error = error.join(", ");
+    }
+
+    return error;
+  }.property("error", "error.@each")
```
mixonic: If not dependent upon a specific property, the best way to do this is will the less expensive `[]` path:

```
}.property("error.[]")
```

tests/acceptance/authentication-test.js
```diff
@@ -251,3 +251,13 @@ test('Creating an account waits on a valid organization name', function() {
     ok(error.length, 'has error on the screen');
   });
 });
+
+test('Creating an account shows validation errors', function() {
+  visit('/signup');
+  click('button:contains(Sign Up)');
```
mixonic: We have a great helper: `clickButton("Sign Up")` will Just Work

mixonic:
I'm going to merge this and make the change to `app/components/validation-feedback/component.js` on master.

Thank you @conrad-vanl!

sandersonet:
Thanks @conrad-vanl !

PR #258, opened 2015-04-02
Add notification messages for unexpected errors.

mixonic:
Using https://github.com/poteto/ember-cli-flash, for example:

Added behind a feature flag: "notifications"


mixonic:
@sandersonet I'm going to say this is for you to review/merge. I think this direction has promise, but it needs more polish. The functionality is behind a feature flag in this PR so it is safe to merge.

PR #260, opened 2015-04-02
Roles index, show, create styles

sandersonet:
cc @mixonic


app/components/permitted-stacks-by-role/component.js
```diff
@@ -0,0 +1,21 @@
...
+        return true;
+      }
+
+      return stack.get('permissions').any((permission) => {
+        if(permission.get('data.links.role') === this.get('role.data.links.self')) {
+          return [this.get('scope'), 'manage'].indexOf(permission.get('scope')) > -1;
+        }
+      });
+    }).map(s => s.get('handle')).join(', ');
+  }.property('scope', 'stacks', 'role.privileged')
```
mixonic: this should probably depend on `stacks.[]` instead of `stacks`. The former observes membership, and the latter only whether the array/object is the same reference.

app/components/permitted-stacks-by-role/component.js
```diff
@@ -0,0 +1,21 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  permittedStacks: function() {
+    return this.get('stacks').filter((stack) => {
```
mixonic: `get` is very cheap, but not free. Additionally, `get` will return undefined if anything in the `complex.path` passed to it is not there.

Additionally we've got a method that handles this logic for you! I suggest:

```javascript
  updatePermittedStacks: function() {
    const role = this.get('role');
    const scope = this.get('scope');
    Ember.RSVP.all(this.get('stacks').map((stack) = > {
      return stack.permitsRole(role, scope).then((permitted) => {
        return permitted && stack;
      });
    })).then((stacks) => {
      const permittedStacks = stacks.filter(r => !!r);
      this.set('permittedStacks', permittedStacks);
    });
  }.observes('scope', 'stacks.[]', 'role.privileged')
```

app/components/user-gravatars/component.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  tagName: 'ul',
+  classNames: 'user-gravatars',
+  truncatedUsers: function() {
+    let count = this.get('count') || this.get('users.length');
+    return this.get('users').slice(0, count);
+  }.property('users.[]')
```
mixonic: missing dependency on `count`

app/organization/roles/index/route.js
```diff
@@ -3,12 +3,23 @@ import Ember from 'ember';
...
+    return Ember.RSVP.hash({
+      roles: organization.get('roles'),
+      stacks: this.store.find('stack')
+    });
+  },
+
+  afterModel(model){
+    return Ember.RSVP.hash({
+      users: Ember.RSVP.all(model.roles.map(r => r.get('users'))),
+      permissions: Ember.RSVP.all(model.stacks.map(s => s.get('permissions')))
```
mixonic: since these two things are not used (just preloaded) it may be worth calling out that they are preloads in a comment.

And maybe even a `// FIXME` comment, since they are such a troll on performance.

sandersonet: Agreed.  We do currently embed permissions on stacks#index, so there shouldn't be any actual requests here.

config/environment.js
```diff
@@ -7,8 +7,8 @@ module.exports = function(environment) {
     baseURL: '/',
     locationType: 'auto',
 
-    authBaseUri: "http://localhost:4000",
-    apiBaseUri: "http://localhost:4001",
+    authBaseUri: "http://auth.aptible.foundry.io",
+    apiBaseUri: "http://api.aptible.foundry.io",
```
mixonic: Whoooooops :-)

tests/acceptance/organization/roles-test.js
```diff
@@ -65,64 +66,8 @@ test(`visiting ${url} shows roles`, (assert) => {
   });
 });
 
-test(`visit ${url} and click to add a user`, (assert) => {
```
mixonic: Why remove this test? How is a user added now?

sandersonet: User management is all from role#show now...

mixonic: ahha! There should be some files for `organization/invite` that can be dropped then, and a router entry to remove.

sandersonet: organization/invite is still linked to from members/index for convenience

mixonic: gotcha.

config/environment.js
sandersonet: Good catch, I updated to use ENV vars where available, so I don't keep trying to sneak this in...

mixonic:
:confetti_ball:

Issue #252, opened 2015-04-01
Unable to accept invitations when not logged in


sandersonet:
If I do not already have an Aptible account, when I click-through an invitation link, I should be prompted to create a user.  This user should immediately be associated to the invited role and added to the organization, skipping Stack/App/PaymentMethod creation along the way.

Currently what happens is that the claim route immediately redirects unauthenticated users to `signup`, where they start the signup process from scratch and are never actually accepted into the organization.

bantic:
discussed w/ @sandersonet 

The signup screen that the unauthenticated user see should be modified to:

  * not show the input for an organization name
  * mention the name of the organization that the user is invited to
  * mention the name of the inviter

Visual:

![image](https://cloud.githubusercontent.com/assets/2023/6948306/f8a3d14e-d87a-11e4-898d-537bbc9eac48.png)

The user can also click the "sign in instead". Either way (after creating a new user, or signing in as an existing user), the next step is to show the user a page that has a prominent button to "accept the invitation to this organization". It should also mention the name of the org and the inviter name.

A logged-in user would go directly to that "accept invitation" page.

PR #263, opened 2015-04-02
Accept invitations when not logged in

bantic:
@mixonic for your review cc @sandersonet 

  * add a /signup/invitations/:invitation_id/:verification_code route
    * this route is for unauthenticated users that click the link they receive in the email when invited to an organization. It has a lot of shared behavior with the current sign up route.
    * extract shared behaviors for signup and signup/invitations into a controller and route mixin

  * the primary differences between signup and signup/invitations are:
    * signup creates an organization after it creates a new user
    * signup redirects to "welcome.first-app" after success
    * signup/invitations does not create an organization
    * signup/invitations redirects to claim/:invitation_id/:verification_code after success
    * signup/invitations shows the name of the organization you are signing up to join

* the primary shared behaviors between the two:
    * validations for user input (email, name, password)
    * create a new user

* change the accept invite page to show the name of the organization the user is invited to join

The auth api was updated in https://github.com/aptible/auth.aptible.com/pull/134 to allow public access to invitations. Rather than trusting the `organization_name` parameter in the URL from the invitation email, we retrieve this info from the auth api.

Also:

  * break the authentication acceptance test file into login, signup/index and signup/invitations tests

The styling and messaging of the new template(s) may need some additional work.

In addition, auth needs to be updated to store the inviter name so that we can display it to the user who is accepting the invitation.


bantic:
fixes #252

app/claim/route.js
```diff
@@ -1,20 +1,43 @@
...
+    controller.set('verificationCode', verificationCode);
+  },
+
+  afterModel(model, transition){
+    if (!this.get('session.isAuthenticated')) {
+      this.get('session').attemptedTransition = transition;
+
+      let invitationId = model.invitation.id;
+      let verificationCode = model.verificationCode;
+      this.transitionTo('signup.invitations', invitationId, verificationCode);
```
mixonic: why transition to the id instead of the model?

app/router.js
```diff
@@ -52,7 +52,9 @@ export default Router.map(function() {
   this.route("stacks");
   this.route("login");
   this.route("logout");
-  this.route("signup");
+  this.route("signup", {}, function(){
+    this.route('invitations', {path:'/invitations/:invitation_id/:verification_code'});
```
mixonic: may make more sense to make this `invitation` even if the path is `invitations`

app/claim/route.js
bantic: as in make an object with an id and verification code?
i.e.: `transitionTo('signup.invitations', {invitationId: id, verificationCode: code}` ? I have no strong preference either way

app/router.js
bantic: sure, that makes sense to me too. I'll fold this in with any other changes

app/claim/route.js
mixonic: Oof, I see. You need access to `params` on that route, which means `model` hook.

app/signup/invitations/route.js
```diff
@@ -0,0 +1,28 @@
...
+    let invitationId = params.invitation_id;
+    let verificationCode = params.verification_code;
+
+    return this.store.find('invitation', invitationId).then( (invitation) => {
+      return { invitation, verificationCode };
+    });
+  },
+
+  setupController: function(controller, model){
+    let {invitation, verificationCode} = model;
```
mixonic: I suggest adding an assertion that `model` is not an invitation instance. It seems easy to trip yourself up on.

The model-hook-skipping behavior really sucks here. IMO assertion is fine, the only other idea I can think of is reading params directly off `transition` inside `beforeModel`.

bantic:
@mixonic addressed your comments. I'll plan on merging this after tests are green unless you have any objections

PR #265, opened 2015-04-03
Errors roles

mixonic:
Adds errors and testing for errors to the roles/show:


mixonic:
Now with a component for showing server validation errors.

Fixes https://github.com/aptible/diesel.aptible.com/issues/248

PR #267, opened 2015-04-03
Track analytics on every page, identify upon sign in

mixonic:
A step forward with https://github.com/aptible/diesel.aptible.com/issues/230. This tracks the page on every page change, as well as moving the identify step into the Torii adapter.

@sandersonet for you to review/merge. I suggest you take a look at what kind of data this generates on staging.

### DIFF
--- a/app/aptible/torii-adapter.js
+++ b/app/aptible/torii-adapter.js
@@ -15,6 +15,7 @@ function persistSession(accessToken){
 }
 
 export default Ember.Object.extend({
+  analytics: Ember.inject.service(),
   fetch: function(){
     var store = this.store;
     return new Ember.RSVP.Promise(function(resolve){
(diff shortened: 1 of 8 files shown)


app/aptible/torii-adapter.js
```diff
@@ -73,5 +79,15 @@ export default Ember.Object.extend({
       clearSession();
       resolve();
     });
+  },
+
+  identifyToAnalytics(user) {
+    const email = user.get('email');
+    this.get('analytics').identify(email, {
```
sandersonet: When we know the ID of the user, this should be ID.

mixonic: So in the `query-box` component we use the email b/c we have no other identifying information yet. Using email elsewhere was just to make sure those people get mapped to real users.

It could be by id, but then I would change the `query-box` component to just send a traits object and let it auto-identify

mixonic: So in the `query-box` component we use the email b/c we have no other 
identifying information yet. Using email elsewhere was just to make sure 
those people get mapped to real users.

It could be by id, but then I would change the `query-box` component to 
just send a traits object and let it auto-identify

sandersonet:
Rebased and merging, thanks @mixonic!

PR #270, opened 2015-04-10
Validate email, app.handle, account.handle, and database.handle for uniqueness during signup

sandersonet:
This PR adds a remote validator. It's used here to validate an attribute's uniqueness, but could easily be used for any kind of remote validation.  This requires a `/claims/:model_type` endpoint that returns a 204 when valid or an error status when invalid (https://github.com/aptible/auth.aptible.com/pull/139)

### DIFF
--- a/app/validators/uniqueness.js
+++ b/app/validators/uniqueness.js
@@ -1,0 +1,45 @@
+import Ember from "ember";
+import Base from 'ember-validations/validators/base';
+import ajax from '../utils/ajax';
+
+export default Base.extend({
+  setupOptions: function() {
+    if(this.options === true) {
+      this.options = {};
+    }
+
+    let options = Ember.$.extend({
+      message: 'is taken.',
+      debounce: 150,
+      type: 'post',
+      data: {},
+    }, this.options);
+
+    this._options = options;
+  }.on('init'),
+
+  getDataPropertyName() {
+    return this.options.paramName || this.property.replace('model.', '');
+  },
+
+  call() {
+    if (!Ember.isBlank(Ember.get(this.model, this.property))) {
+      return Ember.run.debounce(this, this.fetch, this._options.debounce);
+    }
+  },
+
+  fetch() {
+    let options = this._options;
+    let errors = this.errors;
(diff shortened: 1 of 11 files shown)


app/validators/uniqueness.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    let options = Ember.$.extend({
+      message: 'is taken.',
+      debounce: 150,
+      type: 'post',
+      data: {},
+    }, this.options);
+
+    this.set('options', options);
```
mixonic: `options` is not observed or used externally here, so making it `this._options` would probably be ideal.

app/validators/uniqueness.js
```diff
@@ -0,0 +1,57 @@
...
+
+  fetch() {
+    let options = this.get('options');
+    let errors = this.errors;
+    let message = this.options.message;
+    options.data[this.getDataPropertyName()] = this.model.get(this.property);
+
+    // Remove global error handler so it won't catch validation errors
+    let _onerror = Ember.onerror;
+    Ember.onerror = Ember.$.noop;
```
mixonic: Hm, this seems very unnecessary

app/validators/uniqueness.js
```diff
@@ -0,0 +1,57 @@
...
+
+    return new Ember.RSVP.Promise(function(resolve, reject) {
+      errors.clear();
+
+      return Ember.$.ajax(options)
```
mixonic: `Ember.$.ajax` is simply jquery's ajax (with its normal broken promises and lack of runloop). Please use https://github.com/aptible/diesel.aptible.com/blob/master/app/utils/ajax.js#L4

app/validators/uniqueness.js
sandersonet: The application route throws up a flash message when Ember.onerror is triggered.  Is there a better way to prevent that?

sandersonet:
cc @mixonic The automatic `-prod` and `-dev` handle concatenation made the uniqueness validation a little odd, so I've updated this PR to remove that.  Once we have environment settings completed, the double stack creation wouldn't be necessary.

PR #295, opened 2015-05-20
Ember 1.12, Ember-CLI 0.2.5

mixonic:
Some significant upgrades.


config/environment.js
```diff
@@ -61,7 +61,7 @@ module.exports = function(environment) {
     featureFlags: {
       'organization-settings': true,
       'price-estimator': true,
-      'notifications': true
+      'notifications': false
```
mixonic: This feature was unhelpfully squashing errors in master.

tests/acceptance/claim-test.js
```diff
@@ -143,20 +143,3 @@ test('failed verification displays error', function() {
     ok(Ember.$(':contains(error accepting this invitation)').length, 'Failed verifications shows error');
   });
 });
-
-test('failed verification displays error', function() {
```
mixonic: This test is the same as the previous test.

mixonic:
We're blocked by https://github.com/emberjs/ember.js/pull/11225 for the move to Glimmer and by https://github.com/pangratz/ember-cli-htmlbars-inline-precompile/issues/8 for the upgrade to the latest `ember-qunit` (we need a compiler for the new integration tests, one that look [like this](https://github.com/ef4/liquid-fire/blob/master/tests/integration/helpers/liquid-if-test.js#L15)). However this is still great to land.

Issue #291, opened 2015-05-17
Billing Management


sandersonet:
A user should be able to manage their own billing settings

* [x] Billing/Plan [page 310a](https://drive.google.com/file/d/0Bxuaj2I3HLq-WmFsYTZfUFhMcTg/view?usp=sharing) [page 310b](https://drive.google.com/file/d/0Bxuaj2I3HLq-WEhYc0RHR1hUdGs/view?usp=sharing) (PR: #309)
* [x] display billing header (card type, next invoice date, next 4 digits) (PR https://github.com/aptible/dashboard.aptible.com/pull/310)
* [x] Billing/Payment Method [page 311](https://drive.google.com/file/d/0Bxuaj2I3HLq-eUlDZ01xbWVXeDg/view?usp=sharing) (PR: #330)

Other billing screens (not part of this issue):
* Billing/Overview [page 309](https://drive.google.com/file/d/0Bxuaj2I3HLq-UlRkdVpuNzJzYW8/view?usp=sharing)
* Billing/Invoice History [page 312](https://drive.google.com/file/d/0Bxuaj2I3HLq-cERUeDJBVzFIWTQ/view?usp=sharing)

sandersonet:
@samyount @chasballew Feel free to weigh in on these screens and their copy.  

When upgrading, we probably want to throw up a confirmation modal clarifying they'll be charged the $499 immediately.

chasballew:
@sandersonet - The billing can happen immediately, but the stack creation and contract/BAA piece are still async. The contracts are formatted and ready to be triggered through the Echosign API, per #151.

sandersonet:
Good point @chasballew .  Do you have thoughts on how we can word this in the confirmation modal?  Something to the effect of "Once you upgrade, an Aptible Account Manager will contact you with instructions for signing a BAA and using your new PHI-ready environment.  Your card will be charged immediately."  cc @bantic

chasballew:
SGTM

bantic:
updated the modal:

![image](https://cloud.githubusercontent.com/assets/2023/7870415/34c7915c-0557-11e5-8330-18a37158f3bc.png)

sandersonet:
Closing this as finished.  Opening separate tickets for Overview and Invoices screens.

PR #309, opened 2015-05-28
Billing plan

bantic:
Refs #291 

  * [x] add /organizations/:id/billing/plan route
  * [x] style the plan panels on plan page (can we use flexbox?)
  * [ ] ~~display billing overview header (plan name, payment method, invoice date, etc: http://take.ms/kNA55)~~
  * [x] wire the UI to update the organization's plan when requested
  * [x] display a confirmation modal when upgrading
  * [x] Auth: add backend code that handles upgrading an organization (WIP: https://github.com/aptible/auth.aptible.com/pull/151)
  * [x] display error if upgrade fails
  * [x] track an event, show a modal, when a user clicks "ask about our hipaa compliant plans" button
  * [x] change confirmation modal as explained in #291 

@mixonic for your review cc @sandersonet 

@sandersonet : I've used mostly ad-hoc CSS in `app/styles/billing-plans.scss`. Feedback there welcomed and/or you or I can address styling improvements after merge. These seem to be about 80-90% of the way towards matching the mocks. A notable exception is that I didn't fix the height of the panels, so when the window changes they can get out of sync, height-wise. I've added screenshots in comments below of most of the styled UI.

I am going to punt on displaying the billing overview header for now, as it relies on API calls that we do not yet have.

### DIFF
--- a/app/organization/billing/plan/route.js
+++ b/app/organization/billing/plan/route.js
@@ -45,7 +45,7 @@ export default Ember.Route.extend({
                      Ember.get(e, 'responseJSON.message'));
       organization.rollback();
     }).finally(() => {
-      this.controller.set('isUpgrading', false);
+      controller.set('isUpgrading', false);
     });
   },
 
--- a/app/organization/billing/plan/route.js
+++ b/app/organization/billing/plan/route.js
@@ -1,0 +1,52 @@
+import Ember from 'ember';
+import { UPGRADE_PLAN_REQUEST_EVENT } from 'diesel/models/organization';
+
+export default Ember.Route.extend({
+  analytics: Ember.inject.service(),
+  confirmationModal: Ember.inject.service(),
+
+  actions: {
+    upgrade(planType) {
+      this.get('confirmationModal').open({
+        partial: 'confirmation-modals/upgrade-plan',
+        onConfirm: () => this._upgradePlan(planType)
+      });
+    },
+
+    requestUpgrade() {
+      this._trackUpgradeRequest();
+
+      this.get('confirmationModal').open({
+        partial: 'confirmation-modals/request-plan-upgrade'
+      });
+    }
+  },
+
+  _trackUpgradeRequest() {
+      const organization = this.modelFor('organization');
+      const eventName = UPGRADE_PLAN_REQUEST_EVENT;
+      const eventAttributes = {
+        organization_id: organization.get('id'),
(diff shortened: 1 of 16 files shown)


bantic:
Styled UI for platform plan: 
image

bantic:
styled UI when plan = development:

image

bantic:
displayed confirmation modal:

image

bantic:
error message displayed when upgrade fails:

image

bantic:
account upgrade request modal:

image

app/organization/billing/plan/route.js
```diff
@@ -0,0 +1,52 @@
...
+    organization.set('plan', planType);
+
+    controller.set('error', null);
+    controller.set('isUpgrading', true);
+    organization.save().catch((e) => {
+      controller.set('error',
+                     Ember.get(e, 'responseJSON.message'));
+      organization.rollback();
+    }).finally(() => {
+      this.controller.set('isUpgrading', false);
```
mixonic: already plucked it off, and used without `this.` on line 44

app/styles/billing-plans.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  .plan-include-subheader {
+    color: $color-light;
+    font-size: smaller;
+  }
+
+  .plan-btn {
+    width: 100%;
+
+    &[disabled] {
```
mixonic: hm, I would have expected you want the psuedo selector of `&:disabled`, but if the attribute selector works, it works.

mixonic:
looking fantastic. :+1:, will need some UI polish before production.

It would be simple to wrap this in a FF so master isn't blocked from deploy while the styling gets handled.

app/styles/billing-plans.scss
bantic: that seems like a more normative way to do it, I agree. I lifted this from https://github.com/aptible/aptible-sass/blob/c53eb6d47f9877bdac59b144aac3cccf407dcc4c/dist/styles/aptible_buttons.scss#L12. I'll update to use the pseudo selector.

bantic:
Added an FF to hide the "Billing" link in the organization settings sidebar. False in production, true otherwise.

mixonic:
:fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks:

PR #310, opened 2015-05-29
Billing header

bantic:
Adds a Billing header to the organization billing settings pages.

refs #291 

@mixonic for your review

After https://github.com/aptible/ember-cli-aptible-shared/pull/10 is merged, I'll cut a new release and update this PR to use it.

Depends on:
  * [x] ember-cli-aptible-shared pr (https://github.com/aptible/ember-cli-aptible-shared/pull/10)
  * [x] auth pr (https://github.com/aptible/auth.aptible.com/pull/153)

### DIFF
--- a/app/components/billing-header/component.js
+++ b/app/components/billing-header/component.js
@@ -1,0 +1,22 @@
+import Ember from 'ember';
+
+const reads = Ember.computed.reads;
+
+export default Ember.Component.extend({
+  organization: null,
+  billingDetail: null,
+
+  isLoaded: Ember.computed.notEmpty('billingDetail'),
+  plan: reads('organization.plan'),
+  paymentMethodName: reads('billingDetail.paymentMethodName'),
+  paymentMethodDisplay: reads('billingDetail.paymentMethodDisplay'),
+  nextInvoiceDate: reads('billingDetail.nextInvoiceDate'),
+
+  init() {
+    this._super(...arguments);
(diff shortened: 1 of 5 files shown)


bantic:
The header looks like this:

image

app/components/billing-header/component.js
```diff
@@ -0,0 +1,22 @@
...
+  billingDetail: null,
+
+  isLoaded: Ember.computed.notEmpty('billingDetail'),
+  plan: reads('organization.plan'),
+  paymentMethodName: reads('billingDetail.paymentMethodName'),
+  paymentMethodDisplay: reads('billingDetail.paymentMethodDisplay'),
+  nextInvoiceDate: reads('billingDetail.nextInvoiceDate'),
+
+  init() {
+    this._super.apply(this, ...arguments);
```
mixonic: Can also:

```js
  init() {
    this._super(...arguments);
  }
```

It [does the right thing](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&playground=true&code=Component.extend(%7B%0A%20%20init()%20%7B%0A%20%20%20%20this._super(...arguments)%3B%0A%20%20%7D%0A%7D)%3B).

mixonic:
Looks great.

app/components/billing-header/component.js
bantic: :+1: in fact, the way I had it was wrong. Fixing.

Issue #317, opened 2015-06-02
Various regressions from recent refactorings


sandersonet:
The refactoring of aptible-sass and ember-cli-aptible-shared has created several regressions that need fixed before our next production deployment. 

* [x] #319 - Form validation indicators look too much like buttons
* [x] #318 - Flash messages are dismissed on route change
* [x] #314 - Resource tabs no longer indicate the current active tab when redirected to new route
* [x] #304 - Sign up flow looks busted
* [x] #308 - User role changes not immediately reflected
* [x] #300 - Payment info styles busted
* [x] #294 - Vhost buttons are busted
* [x] #293 - Resource header styles are busted
* [x] #289 - Adding a vhost is confusing 
* [x] #287 - Add option for internal vhosts
* [x] #281 - Service container counts break when wrapped
* [x] #321 - Provisioning databases cause broken wrapper
* [x] #323 - Databases should reflect provisioning status
* [x] #324 - Update default vhosts to show virtual_domain rather than external_host
* [x] #325 - Environments index not rendering environments correctly
* [x] #326 - Trigger custom event on newly created stacks

Other minor issues:
* [x] DB selector icons are wack

sandersonet:
Fixed by #327

PR #327, opened 2015-06-04
Regression fixes

sandersonet:
cc @bantic @mixonic This is a monolithic PR that fixes all issues listed in #317.

Requires latest #diesel branch of aptible-sass and https://github.com/aptible/ember-cli-aptible-shared/pull/16


app/app/vhosts/index/controller.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  provisionedDomains: Ember.computed.filter('model', function(vhost) {
+    return vhost.get('isProvisioned');
+  }),
```
bantic: This (and `pendingDomains` and `deprovisionedDomains`) can be even terser: `Ember.computed.filterBy('model', 'isProvisioned')`

app/app/vhosts/new/route.js
```diff
@@ -39,6 +40,8 @@ export default Ember.Route.extend({
         return op.save();
       }).then( () => {
         let message = `Domain ${vhost.get('virtualDomain')} created`;
+        app.get('vhosts').unshiftObject(vhost);
```
bantic: I would think that due to the fact that the `vhost` is created with an app (and app `hasMany('vhost')` and vhost `belongsTo('app')`) that the vhost would automatically be present in the app's `vhosts` property after this vhost is created. I would think that unshifting the vhost directly could result it in being added two times to the app.vhosts property.

If there's a specific ordering of vhosts that you want to ensure, using a CP on the vhost/index controller (creating a property like `sortedVhosts` or similar) may be worthwhile.

app/apps/index/controller.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  deployedApps: Ember.computed.filter('model', function(app) {
+    return app.get('isProvisioned');
+  }),
```
bantic: can use `filterBy` here as well: `Ember.computed.filterBy('model', 'isProvisioned')`

app/components/deploy-instructions/template.hbs
```diff
@@ -0,0 +1,64 @@
...
+
+    <div class="panel-step">
+      <div class="panel-step-num">
+        <h3>2</h3>
+      </div>
+
+      <div class="panel-step-info">
+        <h3>Add a Procfile to your App</h3>
+        <p>
+           Aptible uses your app’s Procfile to configure which services
```
bantic: Not sure if it matters, but that looks like a curly apostrophe (’), and the other one in this template is the straight kind (').

app/router.js
```diff
@@ -6,7 +6,10 @@ var Router = Ember.Router.extend({
   location: config.locationType,
   onTransition: function() {
     this.get('analytics').page();
-    this.get('flashMessages').clearMessages();
+
+    if(config.flashMessageDefaults.sticky) {
```
bantic: Should this be the inverse? I'm not sure what `sticky` means, but I'd expect if flash messages are sticky they would not be cleared by route transitions.

app/app/vhosts/new/route.js
sandersonet: You're right, this unshift is unnecessary with the ED relationships set up correctly.  Removing...

bantic:
:clap: really nicely done. I added a few minor comments.

app/router.js
sandersonet: This code is correct.

The default behavior, `sticky: false`, has the flashMessages disappear after a timeout. This is useful because the messages are visible after a transition, so it's very clear when a new item is created or updated.

However, this timeout breaks the `andThen` behavior in tests, so `sticky` is set to true during testing.

PR #329, opened 2015-06-05
Elasticsearch logdrain dropdown

blakepettersson:
This will add a dropdown for elasticsearch to select all existing elasticsearch databases that a client has, instead of a client having to manually enter a host and port.


blakepettersson:
cc @mixonic @aaw @fancyremarker

blakepettersson:
The test is failing in phantomjs, but is passing with Chrome. My guess is probably because I use `document.createElement()` and phantomjs doesn't support DOM nodes?

app/stack/log-drains/new/route.js
```diff
@@ -2,11 +2,26 @@ import Ember from 'ember';
 import DS from 'ember-data';
 
 export default Ember.Route.extend({
+
   model: function(){
-    return this.store.createRecord('log-drain');
+    var stack = this.modelFor('stack');
+    return Ember.RSVP.all([
```
sandersonet: I think using `Ember.RSVP.hash` is a little better as you can name the promises and then update `setupController` method to use names rather than indexes.  e.g. https://github.com/aptible/dashboard.aptible.com/blob/586767eef70c4b158241e9e9ff4145d6dc9b58d5/app/organization/roles/index/route.js#L7

app/stack/log-drains/new/template.hbs
```diff
@@ -33,14 +33,19 @@
...
+                  <label class="block" for="drain-host">Host</label>
+                {{input name='drain-host' value=model.drainHost class="form-control"}}
+              </div>
+              <div class="">
+                  <label class="block" for="drain-port">Port</label>
+                {{input name='drain-port' value=model.drainPort class="form-control"}}
+              </div>
+          {{else}}
+           {{ database-selector databases=esDatabases select='setDrainFromDatabase' }}
+          {{/if}}
```
sandersonet: Looks like various spacing issues here.  We use two spaces for tabs.

app/stack/log-drains/new/controller.js
```diff
@@ -0,0 +1,26 @@
...
+// We create an anchor since that is, afaik, the easiest way to parse a url in javascript
+function parseUrl(url) {
+    let a = document.createElement('a');
+    a.href = url;
+    return a;
+}
+
+export default Ember.Controller.extend({
+    actions: {
+    setDrainFromDatabase(database) {
```
sandersonet: We have been keeping all action handlers in routes rather than controllers.

sandersonet:
@blakepettersson - Really really solid first contribution, great work!  I commented on some minor issues, @mixonic should also review when he has a chance.

aaw:
:+1: I don't know enough Ember to comment on the implementation but I like the idea of using a dropdown to select the elasticsearch instance!

app/stack/log-drains/new/controller.js
blakepettersson: I tried to move that handler into the route but then I got a whole bunch of errors. I guess that is due to it not having access to the model?

mixonic: Leaving this on the controller feels good to me. It isn't something that persists the model, just an action to modify its state.

app/components/database-selector/template.hbs
```diff
@@ -0,0 +1 @@
+{{ view 'select' content=databases optionLabelPath='content.handle' optionValuePath='content.id' selection=selectedDatabase }}
```
mixonic: This should probably be split into multiple lines.

mixonic:
needs some whitespace/indent cleanup for sure.

Phantom definitely supports creating elements, it is an entire headless browser. It does have bugs, but I'm not familiar with this bad behavior in particular. What I suggest is trying to use Phantom 2.0 instead of 1.9. In `.travis.yml` add these steps to `before_install`:

```
before_install:
  - mkdir phantomjs-2.0.0
  - wget https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -O phantomjs-2.0.0/phantomjs-2.0.0-ubuntu-12.04.tar.bz2
  - tar -xvf phantomjs-2.0.0/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -C phantomjs-2.0.0/
  - export PATH=$PWD/phantomjs-2.0.0:$PATH
```

This is a little slow, but gives us 2.0 today. Let us see what that does.

blakepettersson:
Still failing on phantomjs 2.0...

blakepettersson:
Well, I figured out what the issue is. In phantomjs, `HTMLAnchorElement.password` and `HTMLAnchorElement.username` is `undefined`. Why that is I'm not sure, but I should have a fix for this shortly.

fancyremarker:
See my comment about `drainUsername` and `drainPassword`. Other than that, there are still a few indentation inconsistencies to fix, and one `console.log()` to remove, but if we fix those three things this should be ready to merge.

blakepettersson:
I believe I've fixed the remaining issues, and the tests pass. Is it OK to merge?

fancyremarker:
LGTM. @sandersonet ? See my one comment about `parseUrl`. That can be addressed here or in a separate PR, if you agree.

app/stack/log-drains/new/controller.js
```diff
@@ -0,0 +1,30 @@
...
+  return a;
+}
+
+export default Ember.Controller.extend({
+  actions: {
+    setDrainFromDatabase(database) {
+      let connectionUrl = database.get('connectionUrl');
+      let a = parseUrl(connectionUrl);
+
+      let hostWithoutPort = a.host.substring(0, a.host.lastIndexOf(':')); // Remove port
```
fancyremarker: Should these attributes be set on the object returned by `parseUrl`?

sandersonet: Ya, I think that would clean this method up quite a bit.

sandersonet:
Looks great @blakepettersson, feel free to merge.

PR #330, opened 2015-06-05
Billing update payment

bantic:
Adds a form for updating one's credit card at the organization's billings settings page.

This is ready for review but *do not merge*. ~~It depends on a change to the addon: https://github.com/aptible/ember-cli-aptible-shared/pull/18. After that is merged I'll cut a new release and update the package.json on this PR.~~ (done, -@mixonic)

Also depends on these changes to auth: https://github.com/aptible/auth.aptible.com/pull/156

cc

@mixonic @sandersonet for your review

refs #291 

An important limitation here is that the UI does not immediately reflect the updated billing details. This is due to a bug (?) in ember-data that prevents reloading the `organization`'s `billingDetail` relationship. Calling `reload` on the model does not reuse the HAL `self` link that it had originally used to load itself, it instead constructs a standard `GET /billingDetails/:billing_details_id` url that 404s on the server side.
I've experimented with manually unloading the billingDetail record from the store and reloading it from the organization (as well as reloading the organization and then re-fetching the billingDetail), but ember-data is persistent about not letting that item out of its identity map.

A few options to fix this, none of which is ideal:

  * reload the entire page — the nuclear option
  * add a `GET /billing_details/:billing_details_id` route on the server side so that the reload works
  * modify the HAL adapter to correctly (?) reuse a `self` link when reloading a resource
  * manually fetch the data via ajax (using the `self` link) and `store.push` it back into the store

### DIFF
--- a/app/components/stripe-cc-form/component.js
+++ b/app/components/stripe-cc-form/component.js
@@ -1,0 +1,60 @@
+import Ember from 'ember';
+import { createStripeToken } from 'diesel/utils/stripe';
+
+export default Ember.Component.extend({
+  store: Ember.inject.service(),
+  error: null,
+
+  name: null,
+  number: null,
+  cvc: null,
+  expMonth: null,
+  expYear: null,
+  zip: null,
+
+  isSaving: false,
+
+  organization: null,
+  plan: Ember.computed.reads('organization.plan'),
+
+  actions: {
+    saveCC() {
+      let {
+        name,
+        number,
+        cvc,
+        zip,
+        plan,
+        organization,
+        expMonth: exp_month,
+        expYear: exp_year
+      } = this.getProperties(
+        ['name', 'number', 'cvc',
+         'expMonth', 'expYear', 'zip',
+         'organization', 'plan']);
+
+      let stripeOptions = { name, number, cvc, exp_month, exp_year, zip };
+
+      this.set('isSaving', true);
+      this.set('error', null);
+      let success = false;
+      createStripeToken(stripeOptions).then(({id: stripeToken}) => {
+        const subscription = this.get('store').createRecord(
+          'subscription', { plan, stripeToken, organization }
+        );
+        return subscription.save();
+      }).then(() => {
+        success = true;
+      }).catch((e) => {
+        const message = e.message ? e.message : e;
+        this.set('error', message);
+      }).finally(() => {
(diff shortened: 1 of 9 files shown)


app/components/stripe-cc-form/component.js
```diff
@@ -0,0 +1,60 @@
...
+      createStripeToken(stripeOptions).then(({id: stripeToken}) => {
+        const subscription = this.get('store').createRecord(
+          'subscription', { plan, stripeToken, organization }
+        );
+        return subscription.save();
+      }).then(() => {
+        success = true;
+      }).catch((e) => {
+        const message = e.message ? e.message : e;
+        this.set('error', message);
```
sandersonet: You can use `this.get('flashMessages').danger(message)` instead of this.

sandersonet:
Looks great to me, I defer to you guys on how you want to handle the ED issue.

app/components/stripe-cc-form/component.js
bantic: oh, awesome!

app/components/stripe-cc-form/component.js
```diff
@@ -0,0 +1,60 @@
...
+    saveCC() {
+      let {
+        name,
+        number,
+        cvc,
+        zip,
+        plan,
+        organization,
+        expMonth: exp_month,
+        expYear: exp_year
```
bantic: some new usage of destructuring assignment. left side is property name from the object (the result of the `getProperties` call), right side is local variable name.

app/components/stripe-cc-form/component.js
```diff
@@ -0,0 +1,60 @@
...
+        ['name', 'number', 'cvc',
+         'expMonth', 'expYear', 'zip',
+         'organization', 'plan']);
+
+      let stripeOptions = { name, number, cvc, exp_month, exp_year, zip };
+
+      this.set('isSaving', true);
+      this.set('error', null);
+      let success = false;
+      createStripeToken(stripeOptions).then(({id: stripeToken}) => {
```
bantic: destructuring arguments — stripe responds with a `{id: 'some-token'}` object, this assigns the id property to a local variable `stripeToken` that can be used in the function.

blakepettersson:
Closed in favour of #382

PR #341, opened 2015-06-16
Minor fixes

### LINKED ISSUE #339: Validation errors during signup can cause a user to mistakingly create multiple Stripe subscriptions
https://github.com/aptible/dashboard.aptible.com/blob/master/app/welcome/payment-info/route.js#L32 starts a chain of promises to effectively complete the signup for a user.  If one step of the promises fails, then we assume the entire process has failed and we force the user to start over.  This is a problem if Stripe subscription creation succeeds, but the subsequent stack, app, or db creation fails (for whatever reason).  We should update this route prevent recreation of stripe Subscriptions if the organization already has one.

sandersonet:
This PR hardens the sign up flow with the following changes:

* Stack handle is validated and prevents skipping or moving to payment information when invalid.  Previously, you could leave stack handle blank, or use invalid characters leading to issues like #339
* Prevent multiple stripe subscriptions (and charges) from running when a validation error is generated during signup--fixes #339 
* Minor copy enhancements

### DIFF
--- a/app/app/vhosts/new/template.hbs
+++ b/app/app/vhosts/new/template.hbs
@@ -66,7 +66,7 @@
   </div>
 
   <div class="resource-actions">
-    <button class="btn btn-default" {{action 'cancel'}}>Cancel</button>
-    <button class="btn btn-primary" {{action 'save' model vhostService}}>Save VHost</button>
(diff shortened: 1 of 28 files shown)


sandersonet:
Accidentally pushed to origin.  Deleting the origin branch caused the second failure, merging...

PR #344, opened 2015-06-16
Use "dummy" s3-static-index adapter to statically upload index.html

fancyremarker:
The index.html will be uploaded to the same S3 bucket as other site assets, but this way we get to set no-cache headers on index.html.

cc: @sandersonet

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -1,5 +1,7 @@
 module.exports = {
   production: {
+    buildEnv: 'production',
+
     store: {
       type: 's3-static',
       accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
@@ -18,6 +20,8 @@ module.exports = {
   },
 
   staging: {
+    buildEnv: 'staging',
+
     store: {
       type: 's3-static',
       accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -1,6 +1,11 @@
 module.exports = {
   production: {
     store: {
+      type: 's3-static',
+      accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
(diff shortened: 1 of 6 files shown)


sandersonet:
Looks like ember-deploy defaults build environments to production, which explains why you were seeing  incorrect hosts on aptible-staging.  To fix, I think we need to specify `buildEnv: 'staging'` like so:

```
 staging: {
    buildEnv: 'staging',
    store: {
      type: 's3-static',
      accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
      secretAccessKey: process.env['AWS_SECRET_ACCESS_KEY'],
      acl: 'public-read',
      bucket: 'dashboard.aptible-staging.com'
    },

    assets: {
      type: 's3',
      accessKeyId: process.env['AWS_ACCESS_KEY_ID'],
      secretAccessKey: process.env['AWS_SECRET_ACCESS_KEY'],
      acl: 'public-read',
      bucket: 'dashboard.aptible-staging.com'
    }
  }
```

From their README: https://github.com/ember-cli/ember-cli-deploy#config-file

fancyremarker:
@sandersonet: thanks for that link — I didn't know that you could specify a `buildEnv` explicitly. From observation, I think the configuration may be OK as is; it appears that `buildEnv` is being set from whatever is passed as the `--environment` to `ember deploy`.

sandersonet:
@fancyremarker Hmmm, without specifying `buildEnv` I am seeing the production config defined in index.html when I run `ember deploy -e staging`

fancyremarker:
Hey @sandersonet — you were exactly right. Whatever I thought I had been seeing before was incorrect; explicitly specifying `buildEnv` appears to be necessary for _all_ subtasks: `build`, `deploy:assets`, and `deploy:index`.

PR #345, opened 2015-06-18
Replace AWS keys with those for production-permitted IAM user

fancyremarker:
@sandersonet : This should fix deploying to production bucket (dashboard.aptible.com) on merges to the release branch.


sandersonet:
:+1:

PR #352, opened 2015-07-02
Extract billing

blakepettersson:
Here are the changes required for the dashboard to work with the new billing API. There is still one failing test that needs to be fixed that is directly related to this feature. There are another three failing tests which I do not believe is related to this feature, these failures were introduced in my branch after having merged aptible/dashboard.aptible.com. I will fix these errors in any case, but I'd love to hear some feedback :)


app/organization/billing/plan/route.js
```diff
@@ -34,20 +39,19 @@ export default Ember.Route.extend({
...
     controller.set('isUpgrading', true);
-    organization.save().catch((e) => {
-      Ember.get(this, 'flashMessages').danger(
-        Ember.get(e, 'responseJSON.message')
-      );
 
-      organization.rollback();
+    details.set('plan', planType);
+    details.save().catch((e) => {
+    Ember.get(this, 'flashMessages').danger(
```
sandersonet: Needs tabbed

app/components/billing-header/component.js
```diff
@@ -3,19 +3,18 @@ import Ember from 'ember';
...
   billingDetail: null,
 
   isLoaded: Ember.computed.notEmpty('billingDetail'),
-  plan: reads('organization.plan'),
+  plan: reads('billingDetail.plan'),
   paymentMethodName: reads('billingDetail.paymentMethodName'),
   paymentMethodDisplay: reads('billingDetail.paymentMethodDisplay'),
   nextInvoiceDate: reads('billingDetail.nextInvoiceDate'),
 
   init() {
```
sandersonet: Since you are now loading billingDetail in a model hook, this init() logic is not necessary.  I would just use the organization and billingDetails object that are already loaded in the route.

app/components/billing-header/template.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="layout-container">
     <div class="resource-title">
       <i class="resource-icon resource-org"></i>
-      <h1>{{organization.name}} Billing</h1>
+      <h1>{{organizationName}} Billing</h1>
```
sandersonet: See above, can just be `organization.name`

app/components/billing-header/template.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="layout-container">
     <div class="resource-title">
       <i class="resource-icon resource-org"></i>
-      <h1>{{organization.name}} Billing</h1>
+      <h1>{{billingDetail.organization.name}} Billing</h1>
```
sandersonet: You can remove the `isLoaded` check and property from this component entirely. Since the model is loaded in the route, it will block the rendering of this component entirely until loaded, so the `Loading...` indicator is never going to be visible.

sandersonet:
Great progress @blakepettersson !  In addition to what you have here, you'll also need to account for the transition of billing_contact from auth to billing. That would happen in the route and template here: https://github.com/aptible/dashboard.aptible.com/tree/master/app/organization/contact-settings

blakepettersson:
OK, now that's fixed along with the final failing tests. Does this look good?

app/organization/contact-settings/route.js
```diff
@@ -2,11 +2,25 @@ import Ember from 'ember';
...
+  },
+
   afterModel(model) {
     return Ember.RSVP.all([
-      model.get('users'),
-      model.get('securityOfficer'),
-      model.get('billingContact')
+      model.organization.get('users'),
+      model.organization.get('securityOfficer'),
+      model.organization.get('billingContact')
```
sandersonet: Shouldn't billing contact be pulled from `billingDetail` now, and not organization?

sandersonet:
:clap: @blakepettersson I made one minor comment and the tests still are failing, but otherwise this looks great.  Do you mind squashing all of the WIP commits into a single commit?

blakepettersson:
Now it's rebased and the tests pass! \o/

app/components/operation-item/template.hbs
```diff
@@ -14,5 +14,5 @@
 </div>
 
 <div class="operation-date">
-  on {{format-utc-timestamp operation.createdAt}}
+  on {{format-utc-timestamp operation.createdAt false}}
 </div>
```
sandersonet: Yuck, this is my fault.  We should patch ember-cli-aptible-shared too.

blakepettersson:
The utc-template fix has now been removed, and `ember-aptible-cli-shared` has been bumped to `0.0.24`. 

`app/organization/environments/new/route.js` and the corresponding template is now also using the billing API (I missed that part in my previous commits).

blakepettersson:
Is this ok to merge?

sandersonet:
Good to merge! Thanks @blakepettersson !

PR #353, opened 2015-07-07
Update broccoli-sass to latest.

rwjblue:
Allows usage of iojs 2.2.1 (due to using node-sass^3.0.0).

This is a large jump for node-sass version (from ^0.9.2 to ^3.0.0), and the UX should definitely be reviewed for look/feel issues.  Initial review looks good to me.

@sandersonet for review.


sandersonet:
:+1: Did a quick smoke check and don't see any issues.  Thanks!

PR #354, opened 2015-07-07
Update to ember-cli@0.2.7.

rwjblue:
The blueprinted files that changed upstream have already been overridden, so the only change is for `package.json`.


sandersonet:
:+1:

PR #355, opened 2015-07-07
Update ember-qunit to 0.4.x.

rwjblue:
* Update component unit tests to specify `unit: true` (it is no longer the default).
* Update route unit tests to ensure that `service:analytics` is available.


sandersonet:
:+1:  Thanks!

Issue #346, opened 2015-06-19
Sentry reporting API errors as [object Object]


sandersonet:
https://github.com/damiencaselli/ember-cli-sentry passes errors directly to Raven, which expects a string, not an object.

rwjblue:
Submitted https://github.com/damiencaselli/ember-cli-sentry/pull/10 to cleanup/fix some issues upstream in ember-cli-sentry.

rwjblue:
https://github.com/aptible/dashboard.aptible.com/pull/357 generally addresses this issue (via a number of fixes upstream in Raven and the Raven Ember plugin), but I'm happy to do more testing to confirm this specific scenario is addressed.

sandersonet:
Great fixes, I didn't realize there was already an ember plugin for raven.

I think another issue here is that our APIs return errors as JSON, which [ultimately get passed](https://github.com/getsentry/raven-js/blob/d3e70256b6d1dfcfd20ddae89d0d7acd3403b579/plugins/ember.js#L23) into `raven.captureException` [which treats anything that's not an error, as a string](https://github.com/getsentry/raven-js/blob/d3e70256b6d1dfcfd20ddae89d0d7acd3403b579/src/raven.js#L235).  


An example API error:
```
{
   "code":422,
   "error":"unprocessable_entity",
   "message":"Validation failed: Handle has already been taken"
}
```

rwjblue:
Thank you for the example scenario, that definitely gives me more to test with.

I'll continue to iterate...

rwjblue:
In #357, we added a monkey patch to add a stacktrace in the `[object Object]` case. The hope is that this will help us identify when/why this is happening so we can fix it a bit better.

That monkey patch is deployed to aptible-staging.com at the moment, and I am watching the staging sentry error reports to see if we can trigger this.

frontendherodk:
@rwjblue Hi, do you have update on this?

manjotnms:
Any Update on this issue?

PR #357, opened 2015-07-07
Update sentry version and use custom fork of ember-cli-sentry.

rwjblue:
* Updates Raven to use 1.1.19 (see [changelog](https://github.com/getsentry/raven-js/blob/master/docs/changelog/index.rst) for list of bugfixes), and uses master of ember-cli-sentry to take advantage of upstream Ember plugin fixes.
* Adds a monkey patch to `Raven.captureMessage` to add a `stacktrace` so that we can track down the source of the `[object Object]` errors.

Addresses https://github.com/aptible/dashboard.aptible.com/issues/346.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -4,6 +4,7 @@ import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 import AuthenticatedRouteMixin from 'diesel/mixins/routes/authenticated';
 import { RouteExtension, RouterExtension } from 'diesel/utils/title-route-extensions';
+import monkeyPatchRaven from './ext/raven';
 
 Ember.Route.reopen(RouteExtension);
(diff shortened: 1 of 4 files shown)


sandersonet:
:+1: !

rwjblue:
Just updated to add a monkey patch to `Raven.captureMessage` to add a stacktrace so that we can track down the source of the `[object Object]` errors.

@sandersonet - I'd like to land this, and get it deployed (to at least staging) and see if we can trigger the errors in question.  Any objection?

sandersonet:
Merge at will and it should hit staging shortly thereafter.  Also looks like a minor test failure.

rwjblue:
Whoops, fixed JSHint error and pushed an update.  Will merge once green...

PR #361, opened 2015-07-09
Fix invitation acceptance for unauthenticated existing users

sandersonet:
This PR fixes broken invitation acceptance for existing users who click through an invite link with an unauthenticated session.  The issue was that once the existing user signed in, the verification code was lost and they would then have to click their invitation link again.  This PR fixes that awful UX by storing their claim URL in a cookie and redirecting the user after authentication, effectively removing the extra step of clicking through the invite link again.  Additionally, I've updated the copy to be more friendly, added a larger button for existing users to sign in, and added an ignore button to the claims page.

@chasballew I believe this will fix the issue you've been seeing lately with existing users. 

For @rwjblue to review


chasballew:
Awesome, thanks

rwjblue:
:+1: - LGTM

PR #373, opened 2015-07-15
Update Ember to 1.13.5.

rwjblue:
A few notable changes:

* Change `{{handle-input}}` component tests to integration tests.
* Implement {{database-selector}} without `{{view 'select'}}`.
* Work around setter CP bug in Ember. https://github.com/emberjs/ember.js/pull/11225

---

This PR introduces a number of deprecation notices to the console (952 are triggered when running the full test suite). Should we hold the PR until those are mostly addressed, or merge and work on smaller deprecation PR's after landing the main update?


sandersonet:
:+1: Thanks for this, great work.  I'm fine merging or fixing the deprecations here, completely your call.

app/components/handle-input/component.js
```diff
@@ -36,6 +36,9 @@ export default FocusableInput.extend({
     },
     set(key, value) {
       this._sanitizedValue = sanitizeInput(value);
+
+      this.attrs.value.update(this._sanitizedValue);
```
fancyremarker: Interesting that `get()` / `set()` didn't have to be removed in favor of `didReceiveAttrs()` or another lifecycle hook.

rwjblue: Yes, this was the simplest possible work around to the ember bug we looked at yesterday (linked in description above). All of the other methods I attempted had other downsides. 

In the long run I plan to remove this work around, and move away from two way bindings. Instead the component will just call an action when it's value updates (instead of relying on two way bindings to do the mutation).

The bug is basically that setter CP's do not push their values upstream to mutate properly. This work around is manually calling that upstream update on the attrs object.

rwjblue:
FYI - I've been working on https://github.com/emberjs/ember.js/pull/11833 and https://github.com/emberjs/ember.js/issues/11835 to help with the process of tracking down and dealing with the deprecations.

rwjblue:
Deprecations silenced with [mixonic/ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) along with 874920f. From 32612 deprecations down to 0.

Holding off on merging this until https://github.com/mixonic/ember-cli-deprecation-workflow/issues/16 is resolved (should be soon).

rwjblue:
https://github.com/mixonic/ember-cli-deprecation-workflow/issues/16 is resolved, and a number of the other deprecations have been fixed.

heyjinkim:
https://github.com/heyjinjs/test/issues/1

PR #374, opened 2015-07-15
Ensure database slider is kept in sync.

### LINKED ISSUE #372: Disk size label and slider mismatch when creating new db
Repro:

1. Select a DB type and scale size up. 
2.  hit `Cancel`
3. Select new db

You  will see the slider returns to 10 but the reading on Disk Size does not change.  

image

rwjblue:
* Use `{{diskSize}}` as the `nouislider`'s start value (to ensure they are in sync).
* Use `resetController` to set the controller's state back to the default value when the `database.new` route is exited.

Fixes #372.

@sandersonet -- ready for review.

### DIFF
--- a/app/databases/new/route.js
+++ b/app/databases/new/route.js
@@ -23,6 +23,10 @@ export default Ember.Route.extend({
     }
   },
 
+  resetController() {
+    this.controller.set('diskSize', 10);
+  },
+
   actions: {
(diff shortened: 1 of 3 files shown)


sandersonet:
:clap: Thanks for the quick fix!

wcpines:
Woo!

PR #380, opened 2015-07-20
Render billing contact from the billing API.

blakepettersson:
Fixes an issue where the billing contact couldn't be rendered.


sandersonet:
:+1:

PR #370, opened 2015-07-15
Update ember-cli-aptible-shared to 0.0.25.

rwjblue:
[Changes included between v0.0.21 and v0.0.25](https://github.com/aptible/ember-cli-aptible-shared/compare/6aa0c3d27d830eba7f859ff0925cd92cf3b9b073...v0.0.25)

---

Fixes https://github.com/aptible/dashboard.aptible.com/issues/358.

PR #382, opened 2015-07-21
add {{stripe-cc-form}}

blakepettersson:
This is work that is based from bantic's branch (found in PR #330), now
updated to use the billing API. The issues in that PR are also fixed (the UI does update automagically with this PR).

cc

  * Wire component to stripe and add ability to update billing details
  * Add error and success messaging

### DIFF
--- a/app/components/primitive-select/template.hbs
+++ b/app/components/primitive-select/template.hbs
@@ -1,8 +1,19 @@
-<select class="form-control" name={{name}} {{action 'update' on="change"}}>
-  {{#if prompt}}
-    <option disabled>{{prompt}}</option>
-  {{/if}}
-  {{#each items as |item|}}
-      <option>{{item}}</option>
-  {{/each}}
-</select>
+{{#if data-stripe}}
+    <select class="form-control" {{action 'update' on="change"}} data-stripe={{data-stripe}}>
+      {{#if prompt}}
+        <option disabled>{{prompt}}</option>
+      {{/if}}
+      {{#each items as |item|}}
+          <option>{{item}}</option>
+      {{/each}}
+  </select>
+{{else}}
+    <select class="form-control" {{action 'update' on="change"}} name={{name}}>
+      {{#if prompt}}
+          <option disabled>{{prompt}}</option>
+      {{/if}}
+      {{#each items as |item|}}
+          <option>{{item}}</option>
+      {{/each}}
+    </select>
+{{/if}}
--- a/app/components/primitive-select/template.hbs
+++ b/app/components/primitive-select/template.hbs
@@ -1,19 +1,8 @@
-{{#if data-stripe}}
-    <select class="form-control" {{action 'update' on="change"}} data-stripe={{data-stripe}}>
-      {{#if prompt}}
-        <option disabled>{{prompt}}</option>
-      {{/if}}
-      {{#each items as |item|}}
-          <option>{{item}}</option>
-      {{/each}}
-  </select>
-{{else}}
-    <select class="form-control" {{action 'update' on="change"}} name={{name}}>
-      {{#if prompt}}
-          <option disabled>{{prompt}}</option>
-      {{/if}}
-      {{#each items as |item|}}
-          <option>{{item}}</option>
-      {{/each}}
-    </select>
-{{/if}}
+<select class="form-control" {{action 'update' on="change"}} name={{name}} data-stripe={{data-stripe}}>
+    {{#if prompt}}
+      <option disabled>{{prompt}}</option>
+    {{/if}}
+    {{#each items as |item|}}
+        <option>{{item}}</option>
+    {{/each}}
+</select>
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "core-object": "^1.1.0",
     "ember-cli": "0.2.7",
     "ember-cli-app-version": "0.3.3",
-    "ember-cli-aptible-shared": "0.0.24",
+    "ember-cli-aptible-shared": "0.0.25",
```
rwjblue: This is causing the errors that Travis is having, see https://github.com/aptible/dashboard.aptible.com/pull/370 for confirmation.  The issue is that we are using `Ember.run.later` to poll for updates in provisioning, and that causes extra requests to the server.

blakepettersson: Is this something that is easy to address? Is this something you're working on or should I take a look at it?

rwjblue: I just updated https://github.com/aptible/dashboard.aptible.com/pull/370 to fix the hanging test issue.  I'll merge once green, and then you should be able to rebase this...

blakepettersson:
cc @sandersonet

sandersonet:
Awesome, thanks for this @blake. I already have a follow up ticket for myself to fix the styles here, so I'll do that and enable the billing feature flag as a follow up.

In the mean time, I noticed that all of our stripe form fields still include a name attribute on all of the fields.  (also in the existing [form](https://github.com/aptible/dashboard.aptible.com/blob/master/app/welcome/payment-info/template.hbs) )  We probably should remove all name attributes as that could allow us to accidentally post card information to somewhere it shouldn't go.  Stripe suggests replacing name with `data-stripe`, e.g. `data-stripe="card"`.

Great work!

blakepettersson:
I have replaced the inputs to use `data-stripe` instead of `name`. I'm not sure if the way I did it is the Ember Way ™, but that can be fixed if required. I didn't fix the existing forms yet, but I can do that in a separate PR.

rwjblue:
Just merged https://github.com/aptible/dashboard.aptible.com/pull/373, which changes a few related things here.  I'll send a PR to your branch to update.

blakepettersson:
Wunderbar!

blakepettersson:
Hey @rwjblue, I updated this PR to make the components Ember 2.0 compliant - which means using components instead of using `view`. Can you sanity check?

app/components/primitive-select/template.hbs
```diff
@@ -1,8 +1,19 @@
-<select class="form-control" name={{name}} {{action 'update' on="change"}}>
-  {{#if prompt}}
-    <option disabled>{{prompt}}</option>
-  {{/if}}
-  {{#each items as |item|}}
-      <option>{{item}}</option>
-  {{/each}}
-</select>
+{{#if data-stripe}}
```
rwjblue: I don't think you need this conditional at all, because `<select name={{name}} data-stripe={{data-stripe}}>`  if called without `name` will not include the `name` attribute and without `data-stripe` will not include the `data-stripe` attribute.

app/components/select-month/template.hbs
```diff
@@ -1 +1 @@
-{{primitive-select name=name update=attrs.update prompt=prompt items=months}}
+{{primitive-select name=name data-stripe=data-stripe update=attrs.update prompt=prompt items=months}}
```
rwjblue: Since `{{select-month}}` is a payment only thing ATM, you can omit `name=` here (along with the tweaks mentioned above).

app/components/select-year/template.hbs
```diff
@@ -1 +1 @@
-{{primitive-select name=name items=years update=attrs.update prompt=prompt}}
+{{primitive-select name=name data-stripe=data-stripe items=years update=attrs.update prompt=prompt}}
```
rwjblue: Same as above, I think we can remove `name=` given the other change to primitive select.

app/components/stripe-input/component.js
```diff
@@ -0,0 +1,31 @@
+import Ember from 'ember';
```
rwjblue: Lets call this `one-way-input`
* add both `name` and `data-stripe` to `attributeBindings`
* add a `sanitizeValue` method to the component (that is a noop by default) and have `processValue` call it (roughly like handle-input does)
* update `{{handle-input}}` component to extend from this one but provide its custom `sanitizeValue` function

Doing all of that reduces a bit of duplication (and gives us a single place to make fixes to this common portion).

app/organization/billing/payment-method/template.hbs
```diff
@@ -0,0 +1,24 @@
...
+        <p>
+          {{model.paymentMethodName}} ****{{model.paymentMethodDisplay}}
+          Expires {{model.paymentExpMonth}}/{{model.paymentExpYear}}
+        </p>
+      </div>
+    </div>
+  </div>
+</div>
+
+{{#unless showPaymentForm}}
```
rwjblue: Can this be swapped to the `{{else}}` case of the `{{if}}` just below?

rwjblue:
That last update looks great @blakepettersson!  I left a few comments/tweaks, but they are pretty minor.

rwjblue:
:clap:

PR #393, opened 2015-08-05
Separate certificates from vhosts

sandersonet:
This PR effectively separates SSL certificates from Vhosts and introduces them as a first-class environment resource that can be shared and re-used across apps of the same environment. 

There is now a "Certificates" tab on all environment show pages, where users can create and delete certificates.  Certificates that are currently in use(related to a vhost) are not deletable, and certificate update ability is blocked by the UpdateCertificate items listed [here](https://github.com/aptible/api.aptible.com/issues/140).

screen shot 2015-08-05 at 11 46 53 am

When creating or editing a Vhost, the user can now choose to select from existing certificates (if any exist) or create a new certificate.  If a user chooses to create new (as in a FTUX), the certificate will be saved and will then be re-usable. 

screen shot 2015-08-05 at 11 51 16 am

screen shot 2015-08-05 at 11 51 24 am

Further, this PR deprecates the read/write ability of a VHosts virtual domain and instead extrapolates a common name from the certificate itself.

This depends on https://github.com/aptible/api.aptible.com/pull/228 and https://github.com/aptible/ember-cli-aptible-shared/pull/34


sandersonet:
@rwjblue - Do you mind reviewing?

fancyremarker:
@sandersonet: This should be ready to test on staging. Sweetness is at aptible/sweetness#389 there, and API is at aptible/api.aptible.com#232.

app/app/vhosts/edit/route.js
```diff
@@ -1,55 +1,99 @@
...
+  },
+
+  setupController(controller, model) {
+    let vhost = model.vhost;
+    let certificates = model.certificates;
+
+    controller.set('model', vhost);
+    controller.set('certificates', model.certificates);
+    controller.set('vhostCertificate', certificates.objectAt(0));
+    controller.set('showAddNewCertificate', model.certificates.get('length') === 0);
```
rwjblue: Seems like this should be a CP on the controller.  Perhaps, `Ember.computed.eq`?

app/app/vhosts/edit/route.js
```diff
@@ -1,55 +1,99 @@
...
+      certificate: vhost.get('certificate')
+    });
+  },
+
+  setupController(controller, model) {
+    let vhost = model.vhost;
+    let certificates = model.certificates;
+
+    controller.set('model', vhost);
+    controller.set('certificates', model.certificates);
```
rwjblue: Should this be an `alias` in the controller instead?

app/app/vhosts/edit/route.js
```diff
@@ -1,55 +1,99 @@
...
-      });
-      vhost.save().then( () => {
+        certificatePromise = newCertificate.save();
+      } else {
+        certificatePromise = new Ember.RSVP.resolve(vhost.get('certificate'));
+      }
+
+      certificatePromise.then((certificate) => {
+        vhost.setProperties({ app, certificate });
+        vhost.set('certificateBody', '');
```
rwjblue: Can these extra `vhost.set` calls be moved into the `setProperties` above?

app/app/vhosts/edit/route.js
```diff
@@ -1,55 +1,99 @@
...
-        } else {
-          throw e;
-        }
+        let message = Ember.get(e, 'responseJSON.message') || `There was an error updating ${vhost.get('virtualDomain')}`;
+        Ember.get(this, 'flashMessages').danger(message);
       });
     },
 
-    cancel: function(){
+    showAddNewCertificate(show) {
```
rwjblue: Seems like this should be on the controller (unless I am mistaken this is for controlling UI only).

app/app/vhosts/new/route.js
```diff
@@ -1,37 +1,69 @@
...
     var vhost = model.vhost,
-        services = model.services;
+        services = model.services,
+        certificates = model.certificates;
 
     controller.set('model', vhost);
     controller.set('services', services);
+    controller.set('certificates', certificates);
     controller.set('vhostService', services.objectAt(0));
+    controller.set('vhostCertificate', certificates.objectAt(0));
```
rwjblue: Seems like it should be an `alias`.

app/app/vhosts/new/route.js
```diff
@@ -1,37 +1,69 @@
...
-  setupController: function(controller, model){
+  setupController(controller, model) {
     var vhost = model.vhost,
-        services = model.services;
+        services = model.services,
+        certificates = model.certificates;
 
     controller.set('model', vhost);
     controller.set('services', services);
+    controller.set('certificates', certificates);
```
rwjblue: Seems like it should be an `alias`.

app/app/vhosts/new/route.js
```diff
@@ -1,37 +1,69 @@
...
     controller.set('vhostService', services.objectAt(0));
+    controller.set('vhostCertificate', certificates.objectAt(0));
+    controller.set('showAddNewCertificate', certificates.get('length') === 0);
   },
 
   actions: {
-    willTransition: function(){
+    showAddNewCertificate(show) {
```
rwjblue: Can this be in the controller instead (seems to be for UI only)?

app/app/vhosts/new/template.hbs
```diff
@@ -50,15 +45,34 @@
             </div>
           </div>
 
-          <div class="form-group">
-            <label>Certificate</label>
-            {{textarea class="form-control" value=model.certificate name="certificate" rows=10}}
-          </div>
+          {{#if showAddNewCertificate}}
```
rwjblue: This section seems to be exactly the same between the `new` and `edit` templates, can we tweak this to be a shared component instead?

app/certificates/new/route.js
```diff
@@ -0,0 +1,45 @@
...
+    if (!this.session.get('currentUser.verified')) {
+      controller.set('resourceType', 'app');
+      this.render('unverified');
+    } else {
+      this._super.apply(this, arguments);
+    }
+  },
+
+  actions: {
+    willTransition() {
```
rwjblue: Do we need this?

app/certificates/route.js
```diff
@@ -0,0 +1,31 @@
...
+    return `${stack.get('handle')} Certificates - ${stack.get('organization.name')}`;
+  },
+
+  model() {
+    var stack = this.modelFor('stack');
+    return stack.get('certificates');
+  },
+
+  afterModel(model) {
+    return model.map((certificate) => certificate.get('vhosts'));
```
rwjblue: This seems to return an array of promises, but that doesn't have any "magic" from Ember.  You would need to return `Ember.RSVP.all(model.map(....))`.

rwjblue:
Looks very good!  I left a few inline comments, but feel free to merge at will...

app/app/vhosts/edit/route.js
sandersonet: This is just used to determine the initial view state and isn't always bound.  After initial render, the user can toggle `showAddNewCertificate` with a button click, so I don't think a CP can work here?

rwjblue: Gotcha, thanks for clarifying!

app/certificates/route.js
sandersonet: Doh, meant to do that, nice catch.

PR #397, opened 2015-08-11
Enable organization billing settings for production

sandersonet:
@wcpines - This PR enables the "Billing" tab of organization settings.  This will expose two new helpful features for customers:

1) Customers can now upgrade their plan from development => platform
2) Customers can now update their payment method

Once merged, I'll releases this to production and you can begin directing customers here to update their payment method.


wcpines:
awesome!

PR #403, opened 2015-08-15
Use organization_id instead of organization_url

blakepettersson:
This commit uses organization_id instead of organization_url
to create new stacks with. This is the UI part of aptible/api.aptible.com#231.


sandersonet:
Looks good to me @blakepettersson -- Is it safe to remove the `organizationUrl` attr from shared too?

blakepettersson:
Not completely; there's a method `findStacksFor` in `store.js` which uses the `organizationUrl` to find all stacks by organization.

rwjblue:
LGTM

PR #408, opened 2015-08-21
Minor bug fixes

### LINKED ISSUE #376: UI issue with member remove button when name + org are long
screenshot

### LINKED ISSUE #404: Security Officer should not be removable from an Organization
We should disable the "Remove user from organization" button if the user is in a specialized role like Billing Contact or Security Officer. This will force organizations to first assign a new billing contact or security officer if they wish to remove the user in this role. 

cc @chasballew

### LINKED ISSUE #405: Confusing select box with multiple certificates having the same Common Name
image

cc: @sandersonet 

We can run a query on production to see how much of an actual effect this would have.

### LINKED ISSUE #407: Updating payment method errors are shown as [Object object]
To reproduce, use a declining test card like `4000000000000002`

pasted_image_at_2015_08_19_15_45_360

sandersonet:
Fixes #404
Fixes #405 
Fixes #407 
Fixes #376 

Remove user button when they have a special role:
screenshot 2015-08-21 16 25 41

Certificate select:
screenshot 2015-08-21 16 27 07

Certificate Index:
screenshot 2015-08-21 16 28 20

### DIFF
--- a/app/components/primitive-select/template.hbs
+++ b/app/components/primitive-select/template.hbs
@@ -1,8 +1,8 @@
 <select class="form-control" {{action 'update' on="change"}} name={{name}} data-stripe={{data-stripe}}>
-    {{#if prompt}}
-      <option disabled selected>{{prompt}}</option>
-    {{/if}}
-    {{#each items as |item|}}
-        <option value="{{item}}">{{item}}</option>
-    {{/each}}
+  {{#if prompt}}
+    <option disabled selected={{if (not selected) true null}}>{{prompt}}</option>
+  {{/if}}
+  {{#each items as |item|}}
+    <option selected={{if (eq selected item) true null}}>{{item}}</option>
+  {{/each}}
 </select>
--- a/app/components/primitive-select/template.hbs
+++ b/app/components/primitive-select/template.hbs
@@ -1,8 +1,8 @@
 <select class="form-control" {{action 'update' on="change"}} name={{name}} data-stripe={{data-stripe}}>
     {{#if prompt}}
-      <option disabled>{{prompt}}</option>
+      <option disabled selected>{{prompt}}</option>
     {{/if}}
     {{#each items as |item|}}
-        <option>{{item}}</option>
+        <option value="{{item}}">{{item}}</option>
     {{/each}}
 </select>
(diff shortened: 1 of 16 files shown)


app/components/primitive-select/template.hbs
```diff
@@ -1,8 +1,8 @@
 <select class="form-control" {{action 'update' on="change"}} name={{name}} data-stripe={{data-stripe}}>
     {{#if prompt}}
-      <option disabled>{{prompt}}</option>
+      <option disabled selected>{{prompt}}</option>
```
rwjblue: This is good for our current use cases (it will always select the prompt), but doesn't allow us to have a primitive-select that has an initially selected value.

To do that, the template would look like:

```handlebars
    <select class="form-control" {{action 'update' on="change"}} name={{name}} data-stripe={{data-stripe}}>      
      {{#if prompt}}
        <option disabled selected={{if (not selected) true null}}>{{prompt}}</option>
      {{/if}}
      {{#each items as |item|}}
        <option selected={{if (eq selected item) true null}}>{{item}}</option>
      {{/each}}
    </select>
```

JSBin Demo: http://emberjs.jsbin.com/nayaho/edit?html,js

app/organization/members/edit/controller.js
```diff
@@ -1,5 +1,21 @@
 import Ember from 'ember';
 
 export default Ember.Controller.extend({
-});
+  isSecurityOfficer: Ember.computed('organization.data.links.securityOfficer', 'model.data.links.self', function() {
```
rwjblue: If we find ourselves doing this a bunch, we should make a util computed something like `computedBoundEqual`:

```javascript
// app/macros/bound-equal.js

let { get } = Ember;

export default function(dependentKey, comparisonKey) {
  return Ember.computed(dependentKey, comparisonKey, function(){
    return get(this, dependentKey) === get(this, comparisonKey);
  });
}
```

rwjblue:
:+1:

sandersonet:
Thanks for the review @rwjblue!-- I'll merge now and follow up with another PR for the `computedBoundEqual` recommendation.  I'm pretty sure there are several instances where we would use that.

PR #409, opened 2015-08-23
Preselect current vhost certificate when editing

sandersonet:
This ensures that when editing a VHOST, the currently selected certificate is the default selection option.

Also bumping aptible-shared to fix certificate#apps relation.


rwjblue:
:+1:

PR #414, opened 2015-08-26
Add total estimate to billing header

sandersonet:
This PR adds a total estimated monthly cost to the billing header.  

screenshot 2015-08-26 11 35 54


app/components/total-monthly-estimate/component.js
```diff
@@ -0,0 +1,44 @@
+import Ember from 'ember';
+
+const HOURS_PER_MONTH = 730;
```
rwjblue: This is already specified in https://github.com/aptible/dashboard.aptible.com/blob/c60a1ac7981d166bd4df2493ebf5ea13faab0a23/app/components/usage-quote-by-resource/component.js#L3

Can we have it in one location and import from there?

app/components/total-monthly-estimate/template.hbs
```diff
@@ -0,0 +1 @@
+${{totalDollars}}
```
rwjblue: @gib and I worked on a pretty formatter for currency in a branch on his machine a little while back (with some integration tests), maybe we could bring that in and have `$3,456.78` instead of `$3456.78`?

gib: I'll pick this up (since that branch is still local).

sandersonet: Nice, that would be great!

PR #415, opened 2015-08-26
Format Currency Helper

gib:
My minor addition to the overview work I had done before @sandersonet picked it up. This PR adds a helper to display currency values in cents.

In an effort to display the dollar values consistently, there is some code smell around formatting the `billingDetails.planRate` since it is in dollars. I'm sure there is a better way to provide logic in the helper to do both. Would love to hear thoughts from @sandersonet and @rwjblue on that for a follow up.

### DIFF
--- a/app/components/total-monthly-estimate/component.js
+++ b/app/components/total-monthly-estimate/component.js
@@ -34,6 +34,6 @@ export default Ember.Component.extend({
     return ((netContainers * containerCentsPerHour +
             netDomains * domainCentsPerHour +
             netDisk * diskCentsPerHour) * HOURS_PER_MONTH) +
-            (this.get('billingDetail.planRate') * 100);
+            this.get('billingDetail.planRate');
   })
 });
--- a/app/components/total-monthly-estimate/component.js
+++ b/app/components/total-monthly-estimate/component.js
@@ -24,19 +24,16 @@ export default Ember.Component.extend({
   }),
 
   // Total value
-  total: Ember.computed('netContainers', 'netDomains', 'netDisk', 'billingDetail', function() {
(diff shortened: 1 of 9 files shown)


app/organization/billing/index/template.hbs
```diff
@@ -10,7 +10,7 @@
             Monthly Price:
           </span>
           <span class="usage-value">
-            ${{billingDetail.planRate}}
+            ${{billingDetail.planRate}}.00
```
gib: Would prefer a `format-currency` with an `isDollars` option here.

app/components/total-monthly-estimate/component.js
```diff
@@ -24,19 +24,16 @@ export default Ember.Component.extend({
...
-    return (netContainers * containerCentsPerHour +
+    return ((netContainers * containerCentsPerHour +
             netDomains * domainCentsPerHour +
-            netDisk * diskCentsPerHour) * HOURS_PER_MONTH;
-  }),
-
-  totalDollars: Ember.computed('total', 'billingDetail.planRate', function() {
-    return ((this.get('total') / 100) + this.get('billingDetail.planRate')).toFixed(2);
+            netDisk * diskCentsPerHour) * HOURS_PER_MONTH) +
+            (this.get('billingDetail.planRate') * 100);
```
gib: Dollars to cents on `planRate`. Maybe an update to the `BillingDetail` model is a better place to go.

sandersonet:
@gib -- This is very very nice and definitely DRYs up some duplicated formatting.  I also agree, `billingDetail.planRate` is inconsistent and should probably be updated to return cents so we can remove those special cases.  Want to add that to this PR?

gib:
I'll update!

gib:
@sandersonet, want to give this another look before I merge?

sandersonet:
:+1: Great!

PR #417, opened 2015-08-27
Fix broken Stripe form

### LINKED ISSUE #416: Stripe server-side validations break signup form
Using test card: 4000000000000036 will correctly throw a zip code verification failure on the first attempt.  Subsequent attempts will skip subscription creation [because result will exist from the last failed subscription](https://github.com/aptible/dashboard.aptible.com/blob/master/app/welcome/payment-info/route.js#L49) and attempt to create an `Account` which fails with a 403 "Access Denied" error.

cc @chasballew

Thanks for reporting this issue @dtmunir

sandersonet:
If the first attempt to save a new `billing-detail` fails with a server-side Stripe validation error, all subsequent attempts to validate a payment method will fail.  This is because the subsequent calls to `store.find('billing-detail', organization.get('id'))` will immediately return with the existing `billing-detail` that is still stored in memory.  This will result in subscription creation to immediately resolve rather than attempting to create anew.  Fortunately `authorize! :create, @account` fails on `Account` creation when a `billing-detail` is not present on the organization.

This PR ensures that server-side failures remove the pending `billing-detail` from memory, so a new one will be created.

Fixes #416

### DIFF
--- a/app/welcome/payment-info/route.js
+++ b/app/welcome/payment-info/route.js
@@ -27,8 +27,10 @@ export default Ember.Route.extend({
         address_zip: model.zip
       };
 
-      var organization;
-      var stripeResponse;
+      let organization;
+      let stripeResponse;
+      let pendingSubscription;
+
       saveProgress.set('currentStep', 1);
       Ember.RSVP.hash({
         stripeResponse: createStripeToken(options),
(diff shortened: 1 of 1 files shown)


blakepettersson:
Beautiful! :clap:

rwjblue:
Looks like bower is down (that is causing the Travis failure).

sandersonet:
@rwjblue -- :angry: I guess that leaves me a bit of time to write a spec demonstrating this bug

PR #419, opened 2015-08-28
Sidebar fixes

### LINKED ISSUE #418: Usability issues with adding people to a new account/org
Create a new Aptible account. Wanted to add other engrs to the account. Could figure out how. Here's why:
1) When I hover over my name in the upper right, the entire rectangle highlights.

image

That leads me to believe that the entire area is one big hit spot. I now see that the down triangle gives me a different context menu. Suggestions 1 and 2: Use a clearly separated affordance, like a vertical line between my name and the down triangle. OR make that menu appear always.

Hitting the big button (not the triangle) takes me to my personal account settings, but there's no indication of my domain/company here:

image

Suggestion 3: Put "Tidepool Settings" on the panel to the left.

sandersonet:
I believe this PR along with @gib's aptible/ember-cli-aptible-shared#41 will fix much #418.  Overall I think these are great improvements.

screenshot 2015-08-28 11 31 26

Also included in this PR is a fix to `members/edit/controller.js` that was causing an exception to be thrown when leaving this route.

fixes #418


chasballew:
Very nice.

app/templates/sidebars/settings.hbs
```diff
@@ -1,18 +1,47 @@
...
-    {{#activating-item currentWhen='settings.profile' tagName='li'}}
-      {{link-to 'Profile' 'settings.profile'}}
-    {{/activating-item}}
+    {{#link-to 'organization.members' organization tagName="li"}}
+      {{link-to "Members" "organization.members" organization bubbles=false}}
+    {{/link-to}}
+    {{#link-to 'organization.roles' organization tagName="li"}}
+      {{link-to "Roles" "organization.roles" organization bubbles=false}}
+    {{/link-to}}
+    {{#link-to 'organization.environments' organization tagName="li"}}
```
sandersonet: @rwjblue -- Is there a better way to do this?  We also have a `{{#activating-item}}` component that will activate for the current route path, but it doesn't consider resource IDs, so it won't work here when a user could potentially have multiple organizations.

app/organization/members/index/route.js
```diff
@@ -3,20 +3,24 @@ import Ember from 'ember';
 export default Ember.Route.extend({
   model(){
     let organization = this.modelFor('organization');
-    return organization.get('users');
+
+    return Ember.RSVP.hash({
+      organization: organization,
+      users: organization.get('users')
```
rwjblue: If you make this `model`, you can change the `setupController` to be:

```javascript
setupController(controller, hash) {
  Ember.setProperties(controller, hash);
}
```

Not sure that is enough to worry about though..

app/templates/sidebars/settings.hbs
rwjblue: No, there isn't really a better solution here.  We are working to get `is-active` type helpers in Ember for routing, so you can just do:

```
<li class="{{if (is-active-route "organization.environments" organization) 'active'}}">
  {{! link-to goes here}}
</li>
```

But that is likely at least 2.3/2.4 before landing.

rwjblue: We might be able to make `activiting-item` more intelligent.  I'll look into that separately though...

gib:
Sweet!

PR #420, opened 2015-08-28
Cleans up a weird underline on the user-nav toggles

gib:
See https://github.com/aptible/ember-cli-aptible-shared/pull/41
This will fix that when merged and has no adverse affects until then.


sandersonet:
:+1:

PR #423, opened 2015-08-31
Update to ember-resolver via addon.

rwjblue:
This change will be the default in Ember CLI 2.0.0, just getting a head start...


stefanpenner:
@rwjblue we should include a module at `ember/resolver` to alias back / deprecate or something.

rwjblue:
Sounds good. Will work on it.

rwjblue:
FYI - 'ember/resolver' and 'resolver' module aliases were made and released in 2.0.3.

PR #440, opened 2015-09-22
Ensure correct object is select when prompt is given

rondale-sc:
I was unable to get the test suite to run this test: received the following error:

It should work. :)

### DIFF
--- a/app/components/object-select/component.js
+++ b/app/components/object-select/component.js
@@ -6,7 +6,8 @@ export default Ember.Component.extend({
       const selectedEl = this.$('select')[0];
       const selectedIndex = selectedEl.selectedIndex;
       const items = Ember.A(this.get('items'));
-      const selected = items.objectAt(selectedIndex);
+      const promptOffset = this.get('prompt') ? 1 : 0;
+      const selected = items.objectAt(selectedIndex - promptOffset);
 
       if(selected) {
         this.attrs.update(selected);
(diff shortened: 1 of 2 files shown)


app/components/object-select/component.js
```diff
@@ -1,12 +1,18 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  actions: {
+ actions: {
```
rwjblue: I believe this white space was correct (2 space indent) before, can you double check and update?

app/components/object-select/component.js
```diff
@@ -1,12 +1,18 @@
...
     update: function() {
+      let selected;
       const selectedEl = this.$('select')[0];
       const selectedIndex = selectedEl.selectedIndex;
       const items = Ember.A(this.get('items'));
-      const selected = items.objectAt(selectedIndex);
+
+      if (this.get('prompt')){
+        selected = items.objectAt(selectedIndex - 1);
```
rwjblue: Can you abstract this a little bit?

Something like:

    let promptOffset = this.get('prompt') ? 1 : 0;
    selected = items.objectAt(selectedIndex - promptOffset);

rwjblue:
Thanks for fixing this (and adding a test)!

Left a few minor comments, but this looks great!

rondale-sc:
Updated.

rwjblue:
Thank you!

PR #443, opened 2015-09-23
Fix database header for provisioning databases [fixes #426]

### LINKED ISSUE #426: No ui feedback while db is provisioning
I refreshed the page after provisioning a database and got this:

screen shot 2015-09-04 at 6 11 46 pm

I was confused because I wasn't sure if I was still waiting for it to finish setting up since the `copy` button is there, but there is no url for the database. It still took about 30 seconds of waiting for the url to appear, it would be cool is there was a `pending` notice or something where the url should be.

sandersonet:
Thanks for reporting this issue @evan-007.  This PR updates the database header to only render a spinner while provisioning.  Also added more robust specs.

### DIFF
--- a/app/templates/database/-header.hbs
+++ b/app/templates/database/-header.hbs
@@ -31,31 +31,32 @@
         <h3 class="resource-metadata-value">{{model.type}}</h3>
       </li>
 
-      {{#if model.disk}}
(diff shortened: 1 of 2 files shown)


evan-007:
cool :+1:

PR #455, opened 2015-10-02
Reduce requests

### LINKED ISSUE #452: Redundant requests
There are many redundant requests when loading the dashboard.

As an example, just to load the left sidebar, we load `organizations#index` followed by `organizations#show` for each environment visible in the sidebar--all of which are unnecessary.  

I'll update this ticket after a more thorough investigation of redundant requests and address each in a separate PR.

* [x] Loading individual `organization#show` for each environment listed in sidebar even though they may all share the same organization.
* [x] Getting `accounts#index` followed by `accounts#show` for each stack page
* [x] `accounts#index` is fetched on every `stack#show` page
* [ ] `accounts#show` is fetched on first app, db, or domain sub page

Fixing https://github.com/aptible/ember-cli-aptible-shared/issues/52 will be the biggest win here.

sandersonet:
cc @rwjblue 

This PR significantly reduces the number of requests required to load most pages in Dashboard.  This is done in a couple of ways:

* Load `stacks#index` and `organizations#index` in the root authenticated route
* Avoid loading any resources by ID and instead rely on parent index routes to preload
* Fix `belongsTo` relationships to pull from Store rather than the server [ember-cli-aptible-shared#53](https://github.com/aptible/ember-cli-aptible-shared/pull/53)

fixes #452


sandersonet:
This is deployed to staging.  App feels much faster than production.

Issue #389, opened 2015-07-30
Update billing detail plans for consistency with www.aptible.com


gib:
- development
- platform (PHI-Ready)
- managed (aka production)

fancyremarker:
@gib: Is this specific to the copy on the `/organizations/:id/billing/plan` page?

gib:
Yes. Anywhere we label plans. I originally issued this as something simple I could handle and got diverted.

PR #473, opened 2015-10-15
Updates the billing plan view to match the pricing page on www

gib:
Closes https://github.com/aptible/dashboard.aptible.com/issues/389

There are some interesting things to consider with displaying custom allowances in the “upgrade to a PHI-ready plan!” scenario from a dev plan. The `BillingDetail` model’s `*InPlan` computed properties will
return zero if the current plan is Development (https://github.com/aptible/ember-cli-aptible-shared/blob/master/app/models/billing-detail.js#L67). That’s not what we want to display in this context. The `*Allowance` properties should always be non-zero (https://github.com/aptible/billing.aptible.com/blob/master/app/models/billing_detail.rb#L38) and set to the proper defaults, but just in case, I added a `ternary` helper to avoid `null` and `0` values showing up in
that upgrade plan.

Overkill? Possibly. Poorly named? Also likely… I tried `if-else` and `if-null`, but they felt too specific. 
¯\\\_(ツ)_/¯

Added a helper to display GBs in TBs—to match the pricing page.

screen shot 2015-10-15 at 11 35 26 am


sandersonet:
Nice work @gib!  One minor complication -- this implementation works because we only have 3 plans. However, if we add a 4th Pilot plan, this implementation could get sticky, especially if we have different allowances for Pilot as we do Platform.  The fix would be to only show the billingDetail's allowances if the plan is currently selected. For each plan that isn't selected, allowances can be hard coded to their defaults for that plan.

gib:
We also have those on `BillingDetail`.

gib:
But... we can't access them. Do we expose any globals like
https://github.com/aptible/ember-cli-aptible-shared/blob/master/app/models/billing-detail.js#L16-L35
Or how would we do that.

I'll add a current plan check for now.

app/helpers/ternary.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+
+/**
+ * Returns alt if subject is falsey.
+ */
+export default Ember.Helper.helper(function([ subject, alt ]) {
```
rwjblue: The `ternary` helper confused me initially. Ember already supports ternary like syntax in `{{if someTestCondition truthyValue falseyValue}}`, and this is a bit more like `coalesce` (in SQL terms) where it is returning the first non-truthy value.

Maybe this can be tweaked to something like `find-first`, `first-truthy`, `coalesce`, something better?  Also, we could allow arbitrary param args via:

```javascript
export default Ember.Helper.helper(function(params) {
  return params.find((item) => item);
})
```

gib: I did see that `{{if ... }}` could take three args and use the first arg twice in this case. I think I'll go that route.

sandersonet:
@gib -- Correct, because all of those `const` are exported, you can import them into your controller with:

`import { CONTAINER_ALLOWANCES, DISK_ALLOWANCES, DOMAIN_ALLOWANCES } from 'diesel/models/billing-detail'`;

gib:
plan-type-allowances
Shows how the allowances are used for non-dev plans.

sandersonet:
:+1:

Issue #390, opened 2015-07-31
Pause apps via dashboard (scale to 0)


gib:
This can be done via `aptible-cli` as a work around.
See https://github.com/aptible/aptible-cli/issues/61

sandersonet:
We should probably bump this up in priority now that the API is ready.  @gib, Want to give it a go?

gib:
Sweet! :+1:

PR #491, opened 2015-11-23
Scale containers to 0 in dashboard

gib:
Closes https://github.com/aptible/dashboard.aptible.com/issues/390 with a pretty small change. We should probably follow up with some messaging updates.

Went through creating a new service, which defaults to 1 container (https://github.com/aptible/api.aptible.com/blob/master/app/models/service.rb#L13) and verified the everything saves as expected through the API.

scale-to-zero

### DIFF
--- a/app/components/service-scaler/template.hbs
+++ b/app/components/service-scaler/template.hbs
@@ -41,11 +41,11 @@
         {{no-ui-slider didSlide="setContainerCount"
                        didSet="finishSliding"
                        start=containerCount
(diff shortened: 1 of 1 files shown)


sandersonet:
:+1:

sandersonet:
@fancyremarker -- I think we can ship this once merged?

rwjblue:
:+1:

fancyremarker:
Yup, good to go. Thanks @gib!

PR #492, opened 2015-11-25
Remove global.scss name collision, fixes typekit font rendering

sandersonet:
We have a `_global.scss` in both ember-cli-aptible-shared and dashboard.  One of the updated dependencies in `ember-cliv1.13.10` caused this collision to surface by overriding the aptible-shared _global.scss completely.  

This PR renames the dashboard `_global.scss` so there is no longer a collision and both files are loaded properly.


gib:
:confetti_ball:

PR #494, opened 2015-12-02
enable service scaler price estimator

sandersonet:
This PR enables the pricing estimator located next to all service container sliders.


gib:
I'm excited to see this in action!

PR #500, opened 2015-12-23
Billing overview updates

gib:
Supports dynamic prices for plans. Displays current balance. Adds a “Discounts” section with credit and coupon items if the account has them.

Final nail for https://github.com/aptible/billing.aptible.com/issues/71

It feels like this could be clearer. We've discussed calculating coupons and usage line items for a final total. For now this gets all the information to the user.

@sandersonet


gib:
Merging to review on staging with the billing api updates

wcpines:
This fixes https://github.com/aptible/dashboard.aptible.com/issues/502

PR #501, opened 2015-12-30
Updates ember shared to support log drain status polling

### LINKED ISSUE #483: Dashboard Logdrain status should be updated w/o refresh
Provisioning a papertrail logdrain shows a green spinner and 'provisioning' status, but this remains even after  the API shows `status: provisioned`

Afaik there should be polling/update event. Otherwise a page refresh is necessary.

gib:
Closes #483

Also looks like I missed committing the release tag last week.


gib:
Mergin'

PR #510, opened 2016-01-26
Log Drains - Poll new (for real), restart, and deprovision

gib:
** Early PR with failing tests. Looking for insight from @sandersonet. **

Polling was not previously working right after a new log drain was created because it was "pending" and not "provisioning". This is fixed now but busts acceptance tests.

screen shot 2016-01-25 at 7 09 48 pm

A new component provides a Restart and Deprovision action on provisioned log drains, only the Deprovision action on provisioning / pending log drains, and no actions on a deprovisioning log drain. 

Seeing a deprovisioning log drain in the UI is not likely to happen since sweetness deletes it once the operation is complete and we will remove it from the local store. Subsequent calls to fetch all log drains will no longer include it once the operation is complete.

### DIFF
--- a/app/stack/log-drains/new/route.js
+++ b/app/stack/log-drains/new/route.js
@@ -37,6 +37,7 @@ export default Ember.Route.extend({
 
     createLog(log) {
       log.set('stack', this.modelFor('stack'));
+      log.set('status', 'provisioning'); // Ensures polling for status updates
 
       log.save().then( () => {
         let op = this.store.createRecord('operation', {
--- a/tests/acceptance/log-drains/basic-test.js
+++ b/tests/acceptance/log-drains/basic-test.js
@@ -215,7 +215,7 @@ test(`visit ${addLogUrl} and create log success`, function(assert){
     assert.equal(json.drain_port, drainPort, 'sets drain port');
     assert.equal(json.drain_type, drainType, 'sets drain type');
     assert.equal(json.handle, handle, 'sets drain handle');
-    assert.equal(json.status, status, 'sets darin status');
+    assert.equal(json.status, status, 'sets drain status');
 
     json.id = logDrainId;
     return this.success(json);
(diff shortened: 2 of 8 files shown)


app/stack/log-drains/new/route.js
```diff
@@ -37,6 +37,7 @@ export default Ember.Route.extend({
 
     createLog(log) {
       log.set('stack', this.modelFor('stack'));
+      log.set('status', 'provisioning'); // Ensures polling for status updates
```
gib: This is how I got the polling to work. Or we could added polling on a `pending` status as well in ember-shared.

tests/acceptance/log-drains/basic-test.js
```diff
@@ -294,6 +305,7 @@ test(`visit ${addLogUrl} and create log to elasticsearch`, function(assert){
     assert.equal(json.drain_password, drainPassword);
     assert.equal(json.drain_username, drainUser);
 
+    json.status = 'provistioning';
     json.id = logDrainId;
     return this.success(json);
   });
```
gib: Currently all tests that create a log drain and reload index fail due to the polling
```
Error: Assertion Failed: You cannot reload a record without an ID
```
and 
```
Error: Assertion Failed: You must include an `id` for log-drain in an object passed to `push`
```

tests/acceptance/log-drains/basic-test.js
```diff
@@ -294,6 +305,7 @@ test(`visit ${addLogUrl} and create log to elasticsearch`, function(assert){
     assert.equal(json.drain_password, drainPassword);
     assert.equal(json.drain_username, drainUser);
 
+    json.status = 'provistioning';
```
gib: well... here's a typo

gib:
I'm closing this for a squashed revision with passing tests.

PR #511, opened 2016-01-26
Support HTTPS log drains

krallin:
Requires https://github.com/aptible/api.aptible.com/pull/299 in the backend and https://github.com/aptible/sweetness/pull/497 in Sweetness.

This would be used to deploy Logstash as an app-level Logstash log drain using https://github.com/aptible/docker-logstash/pull/1.

@gib, what do you think? Any recommendations / suggestions?

cc @fancyremarker

### DIFF
--- a/app/stack/log-drains/new/controller.js
+++ b/app/stack/log-drains/new/controller.js
@@ -18,9 +18,13 @@ function parseUrl(url) {
 
 export default Ember.Controller.extend({
   isSyslogDrain: Ember.computed.equal('model.drainType', 'syslog_tls_tcp'),
-  disableSave: Ember.computed('isSyslogDrain', 'esDatabases', function() {
+  isHttpsDrain: Ember.computed.equal('model.drainType', 'https'),
+  isHostPortDrain: Ember.computed('isSyslogDrain', 'isHttpsDrain', function () {
+    return this.get('isSyslogDrain') || this.get('isHttpsDrain');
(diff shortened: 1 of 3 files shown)


gib:
Awesome! I like this a lot. Sorry about the rabbit hole earlier.
You could add a test case to https://github.com/aptible/dashboard.aptible.com/blob/master/tests/acceptance/log-drains/basic-test.js

krallin:
Thanks @gib! 

I added a test that check that submission works and that the port defaults to 443.

gib:
Fantastic! I'm ready to merge...I guess once API and Sweetness are ready.


Thank you @krallin -- thanks for adding the test too!

sandersonet:
This is perfect.  Merge at will.

fancyremarker:
@krallin: Looks like this one [failed its unit test](https://travis-ci.org/aptible/dashboard.aptible.com/builds/107480246#L472) on merge to master, even though it passed on this PR.

krallin:
Did something change between the time I made the PR and we merged it in that could have caused this? (@gib @sandersonet?) If nothing comes to mind, I'll dig for it.

sandersonet:
Not sure why this passed before, but I fixed the broken test here: https://github.com/aptible/dashboard.aptible.com/pull/522

krallin:
Thanks @sandersonet !

Issue #383, opened 2015-07-22
Add ability to remove/restart log drains


sandersonet:
For now:
* [ ] Delete the log drain
* [ ] Create `configure` operation on remaining log drain (https://github.com/aptible/api.aptible.com/blob/master/lib/management_helpers.rb#L55)

Once deprovision/provision operations are supported:
* [ ] Create `deprovision` operation for log drain to delete
* [ ] Create `provision` operation for log drain to create

fancyremarker:
Also, we should provide a way to simply re-configure ("refresh") an existing log drain.

PR #512, opened 2016-01-27
Log Drains - Poll new (for real), restart, and deprovision/delete

gib:
Polling was not previously working right after a new log drain was created because it had a status of "pending" and not "provisioning". The shared polling mixin only acts on "provisioning" objects. This is now fixed...for real.

screen shot 2016-01-25 at 7 09 48 pm

A new component provides a Restart and Deprovision action on provisioned log drains, only the Deprovision action on provisioning / pending log drains, and no actions on a deprovisioning log drain. This is in a tested component.

Seeing a deprovisioning log drain in the UI is not likely to happen since sweetness deletes it once the operation is complete and we now remove it from the local store. Subsequent calls to fetch all log
drains will no longer include it once the operation is complete.

Closes https://github.com/aptible/dashboard.aptible.com/issues/383

### DIFF
--- a/app/components/log-drain-actions/component.js
+++ b/app/components/log-drain-actions/component.js
@@ -14,15 +14,14 @@ export default Ember.Component.extend({
     restart(){
       let message = `Successfully restarted ${this.logDrain.get('handle')}`;
       let errorMessage = `There was an error restarting ${this.logDrain.get('handle')}.`;
-      let logDrain = this.logDrain;
       let component = this;
       var op = this.get('store').createRecord('operation', {
         type: 'configure',
         logDrain: this.logDrain
       });
-      op.save().then(function(){
-        logDrain.set('status', 'provisioning');
-        logDrain.save().then(function() {
+      op.save().then(() => {
+        component.logDrain.set('status', 'provisioning');
+        component.logDrain.save().then(() => {
           component.sendAction('completedAction', message);
         }).catch( (e) => { component.sendError(e, errorMessage); });
       }).catch( (e) => { component.sendError(e, errorMessage); });
@@ -31,19 +30,19 @@ export default Ember.Component.extend({
     deprovision(){
       let message = `Successfully deprovisioned ${this.logDrain.get('handle')}`;
       let errorMessage = `There was an error restarting ${this.logDrain.get('handle')}.`;
-      let logDrain = this.logDrain;
       let component = this;
+      let logDrain = this.logDrain; // Keep a ref for the run later
       var op = this.get('store').createRecord('operation', {
         type: 'deprovision',
         logDrain: logDrain
       });
-      op.save().then(function(){
+      op.save().then(() => {
         logDrain.set('status', 'deprovisioning');
-        logDrain.save().then(function() {
+        logDrain.save().then(() => {
           component.sendAction('completedAction', message);
         }).catch( (e) => { component.sendError(e, errorMessage); });
       }).catch( (e) => { component.sendError(e, errorMessage); });
-      Ember.run.later(component, function() {
+      Ember.run.later(component, () => {
         logDrain.deleteRecord();
       }, 5000);
     }
--- a/app/components/log-drain-actions/component.js
+++ b/app/components/log-drain-actions/component.js
@@ -1,0 +1,51 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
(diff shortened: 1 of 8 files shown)


gib:
@sandersonet - once you give the green light I'm going to do a quick smoke test on staging and push this out.

app/components/log-drain-actions/template.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="panel-heading-actions">
+  {{#unless logDrain.isDeprovisioning}}
+    {{#unless logDrain.isProvisioning}}
+      <button class="btn btn-default btn-xs" {{action "restart" logDrain}}>Restart</button>
+    {{/unless}}
+    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Deprovision</button>
+  {{/unless}}
```
sandersonet: @gib -- Let's move this logic into computed properties.  Nested `#unless` are a bit harder to follow than something like `{{#if showRestartButton}}`.  Computeds are also easier to unit test.

app/stack/log-drains/new/route.js
```diff
@@ -43,6 +43,8 @@ export default Ember.Route.extend({
           type: 'configure',
           logDrain: log
         });
+        // Ensures polling for status updates, no need to save via API
+        log.set('status', 'provisioning');
```
sandersonet: :+1:

app/components/log-drain-actions/component.js
```diff
@@ -0,0 +1,51 @@
...
+        logDrain.save().then(function() {
+          component.sendAction('completedAction', message);
+        }).catch( (e) => { component.sendError(e, errorMessage); });
+      }).catch( (e) => { component.sendError(e, errorMessage); });
+    },
+
+    deprovision(){
+      let message = `Successfully deprovisioned ${this.logDrain.get('handle')}`;
+      let errorMessage = `There was an error restarting ${this.logDrain.get('handle')}.`;
+      let logDrain = this.logDrain;
```
sandersonet: Use this.get('logDrain') throughout.

tests/integration/components/log-drain-actions-test.js
```diff
@@ -0,0 +1,59 @@
...
+});
+
+test('shows no actions for deprovisioning log drains', function(assert) {
+  logDrain.status = 'deprovisioning';
+  logDrain.isProvisioning = false;
+  logDrain.isDeprovisioning = true;
+  this.set('failedAction', failedAction);
+  this.set('completedAction', completedAction);
+  this.set('logDrain', logDrain);
+  this.render(hbs('{{log-drain-actions logDrain=logDrain completedAction="completedAction" failedAction="failedAction"}}'));
```
sandersonet: You can consolidate these:

```
logDrain.setProperties({ status: 'deprovisioning', isProvisioning: false, isDeprovisioning: true })
this.setProperties({ failedAction, completedAction, logDrain }); // Babel allows this syntax for setting properties
```

sandersonet:
@gib -- This is excellent.  There are just some minor comments. Thanks!!

app/components/log-drain-actions/component.js
gib: I created local references for the closures around the `.then`s. Since I have `component`, I can use `component.logDrain`. Maybe I need to try again... `this` had the wrong context.

app/components/log-drain-actions/template.hbs
gib: :+1:

app/components/log-drain-actions/template.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="panel-heading-actions">
+  {{#if logDrain.isProvisioned}}
+    <button class="btn btn-default btn-xs" {{action "restart" logDrain}}>Restart</button>
+  {{/if}}
+  {{#unless logDrain.hasBeenDeprovisioned}}
+    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Deprovision</button>
+  {{/unless}}
```
gib: That nested logic really wasn't necessary -- thanks for that note. It actually allowed restarts of `pending` log drains, which wasn't intended.

It's now correctly showing the "Restart" action _only_ for `provisioned` log drains and the "Deprovision" action for anything other than `deprovisioning` (or `deprovisioned`) log drains.

Issue #514, opened 2016-01-29
Add a confirm to deprovision a log drain


gib:
Thanks @blakepettersson for pointing this out. Pretty major oversight... demos FTW!

PR #515, opened 2016-01-29
Confirm deletes of log drains and certificates

gib:
Closes https://github.com/aptible/dashboard.aptible.com/issues/514

confirm-delete

We haven't used these native confirm dialogs in the app before. It may be an Ember anti-pattern, but I think it's a good compromise to get the log drain actions available. (@sandersonet, @mixonic)

### DIFF
--- a/app/components/log-drain-actions/template.hbs
+++ b/app/components/log-drain-actions/template.hbs
@@ -3,6 +3,6 @@
     <button class="btn btn-default btn-xs" {{action "restart" logDrain}}>Restart</button>
   {{/if}}
   {{#unless logDrain.hasBeenDeprovisioned}}
-    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Deprovision</button>
+    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Delete</button>
   {{/unless}}
 </div>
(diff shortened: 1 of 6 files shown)


app/components/log-drain-actions/template.hbs
```diff
@@ -3,6 +3,6 @@
     <button class="btn btn-default btn-xs" {{action "restart" logDrain}}>Restart</button>
   {{/if}}
   {{#unless logDrain.hasBeenDeprovisioned}}
-    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Deprovision</button>
+    <button class="btn btn-default btn-xs" {{action "deprovision" logDrain}}>Delete</button>
```
gib: This does a better job of setting expectations. It gets deprovisioned and then deleted via the operation.

tests/acceptance/certificates/delete-test.js
```diff
@@ -79,6 +79,8 @@ test(`visiting ${url} and deleting a certificate should delete the certificate`,
   signInAndVisit(url);
   andThen(function() {
     let deleteButton = find('button:contains(Delete)');
+    // Clobber window confirm to accept delete.
+    window.confirm = () => { return true; };
```
gib: This is my leeroy solution to keep the test runner happy... overwrite the native window method to return true without displaying the test-runner-blocking dialog.

mixonic: I suggest doing this globally for the whole test suite. For example as we do here: https://github.com/aptible/dashboard.aptible.com/blob/master/tests/test-helper.js#L18

Most significantly setting it here does leak it into other tests.

PR #524, opened 2016-02-11
`--initialize-from` support

krallin:
This adds basic support for `--initialize-from` in the UI

Specifically, it:

+ Adds a "Replicate" tab to databases that support it. The tab currently prompts users to contact support, and shows replicas after we've provisioned them.
+ Updates the "Deprovision" tab to not let you delete a database unless you've deleted all dependents first. This is mainly here to ensure users don't expect to just be able to just use the replica after deleting the master.

Here are a couple of screenshots:

- https://www.evernote.com/l/AXs21987ukdOzrVMUuNFp5JsEhbcC-7m_V0
- https://www.evernote.com/l/AXvxQznQZL5OBJKBPGH3d6JQnXgDWZSIB8E

In the future (e.g. when we support ElasticSearch replication), we'll need to make this a little smarter and have the names depend on the DB type (e.g. "cluster" instead of "replicate).

Thanks @sandersonet for helping me understand where the issue fixed in https://github.com/201-created/ember-data-hal-9000/pull/40 was coming from :)

Note: this depends on https://github.com/aptible/ember-cli-aptible-shared/pull/84 and https://github.com/201-created/ember-data-hal-9000/pull/40

### DIFF
--- a/app/components/operation-icon/component.js
+++ b/app/components/operation-icon/component.js
@@ -11,7 +11,8 @@ var types = {
   tunnel: 'fa-exchange',
   clone: 'fa-copy',
   deprovision: 'fa-remove',
-  reprovision: 'fa-refresh'
+  reprovision: 'fa-refresh',
+  replicate: 'fa-clone'
 };
 
 var statuses = {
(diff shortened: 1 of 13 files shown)


krallin:
Note: the tests are passing locally, but of course they won't pass in Travis until the deps are fixed.

sandersonet:
Perfect! Merge at will.

PR #529, opened 2016-02-24
Support clustering

krallin:
Add support for clustered databases (currently only MongoDB). This works similarly to replicated databases, but with the little tweak that you can't daisy-chain clustered databases. 

As noted in the commit messages, that's not actually *needed* (the backend will accept the daisy-chaining just fine and you'll *still* have a cluster), but it'll avoid confusing UX where you don't see all the members of a cluster because you happened to have daisy-chained them.


krallin:
Note: CI will fail until https://github.com/aptible/ember-cli-aptible-shared/pull/90 is merged in

app/templates/database/-replicate.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="panel-heading">
+  <h3>Submit a support request</h3>
+</div>
+<div class="panel-body">
+  <p>
+  Replication and clustering are currently in beta.
+  To add a new replica or cluster member,
+  <a href="{{supportPortal}}">please contact support</a>.
```
sandersonet: We have a helper that encapsulates this, you can see it in action here: https://github.com/aptible/dashboard.aptible.com/blob/e2df50c16c71733f97eb7ef57a0147ac8ab46627/app/stack/activate/template.hbs#L8

sandersonet:
:+1:  Awesome, thanks for this!  Just one minor comment that should clean some things up.  Merge/release at will!

gib:
Nice!

krallin:
Oh; this is nice! I didn't know about this in #524!

Does the fix in my last commit work for you? :)

sandersonet:
@krallin -- Just realized that we are sunsetting support.aptible.com/contact in favor of contact.aptible.com.  Can you update the link-to-aptible to use cotnact.aptible.com by adding that to the `aptibleHosts` block in config/environment.js? Thanks!

krallin:
@sandersonet amended the last commit; is that what you had in mind?

sandersonet:
Yep, that's perfect.  Thanks @krallin

krallin:
Awesome; thanks @sandersonet. Should I merge now so we can play with it on staging?

@fancyremarker did you want to have a quick look first?

sandersonet:
Do it

fancyremarker:
My only concern here is that the knowledge of which databases are "clustered" versus "replicated" is known only to Dashboard. Do we think this will create complexity going forward? Perhaps we could consider the specific example of adding replicate/cluster functionality to aptible-cli.

krallin:
Notes from our chat in Slack: 

- The differentiation is mainly here to improve the UX in Dashboard, but not needed for the backend to work properly.
- It's fine for `aptible-cli` to mirror the underlying low-level `replicate` operation.

I'll merge this as-is for now, and of course this is something we could revisit if we feel it's worth investing  time in a proper "cluster" concept on the backend!

Issue #516, opened 2016-02-01
Expose container_memory_limit_mb from UI


blakepettersson:
Add ability to set the `container_memory_limit_mb` directly from the UI. Maybe we should have some kind of slider to set this on a service level?

fancyremarker:
:+1: I think that makes sense. The options should probably be limited to: 256 MB, 512 MB, 1 GB, 2 GB, 4 GB, 8 GB.

sandersonet:
@fancyremarker Do we want to expose this for both DB and app containers?

fancyremarker:
@sandersonet Yes, I think we'd want it for both, at least eventually. One catch is that updating the `container_memory_limit_mb` _won't_ immediately change the memory limit for live app or database. As of right now, the customer would need to run an `aptible restart` (for an app) or ask us to restart (for a database) in order for the change to take effect. There are a couple ways to improve this experience going forward:

* In Sweetness, either (a) add a `ResizeService` operation, or (b) allow `RestartApp` and `RestartDatabase` to take `container_memory_limit_mb` as an argument. These operations would be the only way to update the value of `container_memory_limit_mb`. In other words, it would not be directly updatable. This has the advantage that it's least confusing.
* In Dashboard, we could list the current containers and their memory limits under each service. A customer would _still_ need to restart their app in this scenario. This has the advantage that, once our billing summary reflects container sizing, we could make it easier for customers to tweak app/database footprints, see how their bill would change, and then actually restart their apps/databases in order to effect the change.

blakepettersson:
@fancyremarker regarding what needs to be done in Sweetness - couldn't resizing a container be considered a `ScaleService` operation? In a way, the only real difference is that they are scaling the size of the container rather than scaling the number of containers.

fancyremarker:
@blakepettersson: Yes, I agree with you. I think using `ScaleService` makes good sense here, instead of adding a `ResizeService` operation.

blakepettersson:
Done!

PR #534, opened 2016-02-27
Add slider for service container size

blakepettersson:
Initial commit for setting the service container size for an app.

* Needs some extra styling
* Needs some extra test cases
* This has not been wired up to set the container size on the api yet
* `Object.assign` is not included in PhantomJS 1.9.x, so the tests
there currently fail. Not sure how to handle that.

Relates to #516.

cc @sandersonet @gib

### DIFF
--- a/app/app/services/index/route.js
+++ b/app/app/services/index/route.js
@@ -20,7 +20,7 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    scaleService: function(service, containerCount, deferred){
+    scaleService: function(service, containerCount, containerSize, deferred) {
       let reloadUntilOperationStatusChanged = (operation, maximumTimeout, timeout) => {
         return operation.reload().then((o) => {
           return new Ember.RSVP.Promise((resolve, reject) => {
@@ -44,6 +44,7 @@ export default Ember.Route.extend({
 
       this.store.createRecord('operation', {
         type: 'scale',
+        containerSize: containerSize,
         containerCount: containerCount,
         service: service
       }).save()
@@ -52,5 +53,4 @@ export default Ember.Route.extend({
         .then(deferred.resolve, deferred.reject);
     }
   }
-
 });
--- a/app/components/estimated-cost/component.js
+++ b/app/components/estimated-cost/component.js
@@ -22,7 +22,7 @@ export default Ember.Component.extend({
 
   total: function() {
     return centsToDollars(
-      hoursPerMonth * this.get('count') * this.get('stack.appContainerCentsPerHour')
+      hoursPerMonth * (this.get('count') * (this.get('size') / 1024)) * this.get('stack.appContainerCentsPerHour')
     );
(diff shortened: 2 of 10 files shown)


blakepettersson:
Pending the merges of aptible/api.aptible.com#316 and aptible/ember-cli-aptible-shared#91, it now sets the `containerSize` on the `Operation` model. We need to reload the `Service` model if `operation.save()` is successful though, would `service.reload()` be enough?

sandersonet:
@blakepettersson -- I think you can replace the `Object.assign` calls with `Ember.$.extend()` to the same affect.

Yep `service.reload()` after the operation saves should be sufficient.  

Would you want to take a stab at also updating the billing overview screen in a follow-up PR? 

Looks great, thanks for this!

blakepettersson:
I found `Ember.merge` which does the same thing as `Object.assign`. 

I'll take a look at the billing overview after merging this PR!

blakepettersson:
I added the ability to disable the slider if the service is on a Sweetness v1 stack.

blakepettersson:
I added some styling for when `.noUi-connect` is disabled. The sliders themselves look a bit wonky, but that is easily solved by wrapping each `container-count` and `slider-wrapper` in a div that's styled with 

```clear: left; 
   height: 50px;
```

I'd have added that myself but I'm not sure what the conventions are for class names etc.

sandersonet:
> I'd have added that myself but I'm not sure what the conventions are for class names etc.

Our general methodology has been:  Styles should either be specific enough that they can belong in a component's css file `app/styles/components/*` or general enough that they can be added to `aptible-shared` in `app/styles/*`.   Since this specifically affects sliders, I'd just add a new class here: https://github.com/aptible/dashboard.aptible.com/blob/master/app/styles/components/slider.scss with those styles.

blakepettersson:
Alright, there should be no more outstanding UI issues. 

Good to :shipit: @sandersonet @gib? 

PS: what do you guys use to create GIF screen captures?

chasballew:
I use http://recordit.co/, @mixonic uses http://www.cockos.com/licecap/

sandersonet:
:+1:  Looks great, thanks @blakepettersson

mixonic:
Monosnap also records videos now :-) but the gif export is poor quality

gib:
I'm a LiceCap fan too!

blakepettersson:
I settled on using Licecap for the future 😄

blakepettersson:
I added a label to make it more obvious that the lower slider scales memory, as well as changing "Saving" to "Scaling" per the suggestion of @krallin. So the final result can be seen below.



scaler-take-2

krallin:
❤️

sandersonet:
:+1:  I'm also wondering if we should disable the sliders and the button while `Scaling...`?

gib:
👏

blakepettersson:
Boom

scaler-take-3

Issue #496, opened 2015-12-03
Domains listed in environment metadata don't update correctly when changing stacks


sandersonet:

gib:
This is still an issue. Easily reproduced on production. The totals are correct, but the listed domains are not. Thanks for the find @chasballew.

chasballew:
I found it

![](http://www.summerbeats.com.br/img/golden-ticket.gif)

gib:
I love this

PR #536, opened 2016-02-29
Fix stale environment header vhost metadata

gib:
**Really** closes https://github.com/aptible/dashboard.aptible.com/issues/496 by binding the computed vhosts metadata to the parent environment.

Here's the bug, requiring a refresh to get the right data...
bug

With the parent model included in the ember computed scope, the update works as expected...
fixed


gib:
@sandersonet, I'm not sure why Travis isn't picking this up, but the tests are passing locally. The fix looks obvious now, but took me a bit to set up locally.

sandersonet:
:+1: I'm fine merging this even though Travis be trollin'

gib:
:joy: Thanks!

PR #545, opened 2016-03-11
Impersonation improvements

krallin:
A few things here:

- Fist, fixed a typo I introduced in the last impersonation PR.
- Second, changed how we redirect to the index after impersonation (to be consistent with how login and logout do it). In practice I noticed the `transitionTo` redirect wasn't always succeeding, which according to the network logs was usually because the app was trying to access data it knew existed (because the store hadn't been cleared after changing users), but was no longer able to access (because its user no longer had privileged).
- Third, Added support for organization-href impersonation, which delegates finding who you want to impersonate to auth.aptible.com: you just tell it the organization href, and you'll get logged in as a privileged user within that organization (though it will sometimes fail to find such a user until https://github.com/aptible/auth.aptible.com/pull/206 is merged in).

--

cc @gib @sandersonet for review

cc @wcpines: this should fit your workflow better. Once you have located the organization, just dump its href with https://github.com/aptible/dangerzone/pull/61, and use that to impersonate.

### DIFF
--- a/app/settings/impersonate/route.js
+++ b/app/settings/impersonate/route.js
@@ -44,7 +44,7 @@ export default Ember.Route.extend({
       .then(() => {
         return this.transitionTo('index');
       }, (e) => {
-        this.currentModel.set('error', `Error: ${e.message || JSON.stringify()}`);
+        this.currentModel.set('error', `Error: ${e.message || JSON.stringify(e)}`);
       })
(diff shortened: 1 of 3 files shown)


wcpines:
:clap:

sandersonet:
:+1: Perfect

PR #547, opened 2016-03-11
fix infinite redirect for 'pending' apps with multiple services

sandersonet:
Thanks for reporting this @krallin


krallin:
That was fast! :tada:

PR #551, opened 2016-03-16
Create and delete default vhosts

blakepettersson:
default-vhost-from-ui
This commit adds the ability to create and delete default vhosts from
the dashboard.

This depends on aptible/api.aptible.com#307 and aptible/sweetness#525
and hence this cannot be merged until those PRs are merged.

As for the UX itself, the radio button is only revealed if the service
does not already have a default vhost set on it. An outstanding
question is how to only display this option for stacks that have
default vhosts enabled (there are a bunch of production stacks which
have this enabled as well as all(?) shared stacks). Or should this be
enabled for all stacks?

Is the radio button clear enough for customers to know what it's
about? Is there any way it could be made clearer?

cc @gib @sandersonet @fancyremarker


sandersonet:
> Is the radio button clear enough for customers to know what it's about? Is there any way it could be made clearer?

I think it could be a bit more clear.  Can't we reliably predict the external host for default domains?  If so, the options could be:

#### Domain Name

* [ ] Use `app-123.aptible.in` default domain name
* [ ] Use a custom domain name

gib:
@blakepettersson! :clap: 

Once the API is re-deployed and I merge that drag and drop textarea, we could use those here.

app/components/create-vhost/component.js
```diff
@@ -0,0 +1,40 @@
...
+  }.observes('vhostService'),
+
+  setHasDefaultVhost: function(vhostService) {
+    if(vhostService) {
+      vhostService.get('vhosts').then((vhosts) => {
+        let filtered = vhosts.filter((vhost) => vhost.get('isDefault'));
+        this.set('hasDefaultVhost', filtered.length > 0);
+      });
+    }
+  },
```
sandersonet: I think you can collapse `setInitalVhostServiceState`, `vhostServiceObserver`, and `setHasDefautlVhost` into a single computed property:


```javascript

 hasDefaultVhost: Ember.computed('vhostService', 'vhostService.vhosts.@each.isDefault', function() {
  let vhostService = this.get('vhostService');
  let hasDefault = false;
  if (vhostService && vhostService.get('vhosts').length > 0) {
    hasDefault = vhostService.get('vhosts').filter((vhost) => vhost.get('isDefault')).length > 0
  }
  return hasDefault;
 })
```

fancyremarker:
:+1: Thanks @blakepettersson ! 

I agree with @sandersonet 's suggestion for clarifying the choices. Also:

> An outstanding question is how to only display this option for stacks that have default vhosts enabled [...] Or should this be enabled for all stacks?

Let's enable it for all stacks!

Are aptible/api.aptible.com#307 and aptible/sweetness#525 ready for final review?

app/components/create-vhost/component.js
blakepettersson: Hmm, that's doesn't work for me sadly.. the state is always false with this, even for services where I know a default vhost exists.

blakepettersson:
@fancyremarker 

> I agree with @sandersonet 's suggestion for clarifying the choices

Done.

> Let's enable it for all stacks!

:tada: 

> Are aptible/api.aptible.com#307 and aptible/sweetness#525 ready for final review?

Yes!

fancyremarker:
OK, looks great. Is the following rollout schedule correct? Am I missing anything?

* [ ] Fix merge conflict in aptible/api.aptible.com#307
* [ ] Fix `Vhost.where(default: false, certificate_id: nil)` on staging/production (as I mention on aptible/api.aptible.com#307)
* [ ] Merge aptible/sweetness#525, then aptible/api.aptible.com#307, then this PR
* [ ] Test on staging
* [ ] Release Sweetness, API, and Dashboard in the same order

blakepettersson:
> Release Sweetness, API, and Dashboard in the same order

I believe the order would be API, Sweetness and Dashboard. We would also need to set the default certificate and key on all stacks that don't already have that enabled.

blakepettersson:
Alright, is this good to merge? 

sandersonet:
Looks good to me

fancyremarker:
@blakepettersson : Yes sir. I'd just stick to the order in the checklist above. (I've edited it to include your correction.)

PR #552, opened 2016-03-16
Frame Buster

### LINKED ISSUE #542: Bust iframes

gib:
Closes #542

Been working on a larger task, but saw a quick path here and decided to knock it out. `bustFrames` blocks pages from sourcing dashboard in a frame/iframe tag. I smoke tested this locally, but couldn’t think of a reasonable way to include a test in the repo without some shenanigans.

Bust some frames Tony!

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -4,6 +4,7 @@ import loadInitializers from 'ember/load-initializers';
 import config from './config/environment';
 import { RouteExtension, RouterExtension } from 'diesel/utils/title-route-extensions';
 import monkeyPatchRaven from './ext/raven';
+import bustFrames from './ext/bust_frames';
 
 Ember.Route.reopen(RouteExtension);
 Ember.Router.reopen(RouterExtension);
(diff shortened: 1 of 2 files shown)


sandersonet:
:+1:

krallin:
Should we / can we perhaps set the [`X-Frame-Options`](https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options) header on dashboard? 

If we're serving from S3 / CF using static hosting (I believe we are, right?), I seem to remember it's possible to set headers on the bucket objects.

sandersonet:
@krallin is exactly right.  We should be able to just set `X-Frame-Options` on `index.html` by updating our deploy script here: https://github.com/aptible/dashboard.aptible.com/blob/master/lib/ember-deploy-s3-static-index/adapter.js#L42

gib:
Bummer. Looks like this is not supported: https://forums.aws.amazon.com/thread.jspa?threadID=149569
I'll keep looking.

gib:
Closing here... added to `ember-cli-aptible-shared` https://github.com/aptible/ember-cli-aptible-shared/pull/101

PR #553, opened 2016-03-16
Let dashboard show container memory metrics

krallin:
Note: this is **not** ready for merging (~~the URL for the metrics service is wrong and hardcoded, and~~ there's a commit that points dashboard to my clone of ember-cli-aptible-shared that needs to be removed after https://github.com/aptible/ember-cli-aptible-shared/pull/99 is released)

---

Now, although this isn't ready for merging, I'd like to start getting some feedback on this if possible (specifics below).

In a nutshell, this exposes container memory metrics in dashboard. Here are a couple screenshots:

image

image

image

Here's how this works:

- Metrics are collected using [cadvisor](github.com/google/cadvisor) and pushed to an influxdb instance.
- A proxy service (the `statsProxy` you can see in `environment.js`) is hosted on Amazon Lambda. It can receive a request for metrics for a set of container IDs, and will send back formatted metrics for Dashboard to use.
- Dashboard consumes those metrics by submitting a set of container IDs to the proxy service, and displaying the results using c3.js.

Now, here's what's included in this PR:

- A controller mixin, a route mixin, and a template partial to render the metrics page.
- Those are used in new routes for app and DB metrics.
- An Ember component  that renders a chart using c3 (the options there are limited to what I needed right now — note that there were a few packages integrating c3 and ember; I tried those first, but they didn't work too well).

A few things I'd like to call out for this discussion:

- This includes d3 and c3 as dependencies for Dashboard. Are those dependencies too big to be reasonable to include?
- This is all imported in one commit. Do we want one for the c3-chart component and one for the rest?
- There's a lot of duplication between the app and DB tests. Do we want to change that?
- This makes requests to the metrics service via the controller. I think I remember seeing somewhere that we wanted to make requests in routes, but I'm not sure. Let me know if I'm imagining things here!

---

cc @sandersonet @gib @fancyremarker


sandersonet:
Wow! @krallin, this is really really great.  

To answer some of your questions:

> This includes d3 and c3 as dependencies for Dashboard. Are those dependencies too big to be reasonable to include?

I'm fine with it

> This is all imported in one commit. Do we want one for the c3-chart component and one for the rest?

I'm good with how it is.

> There's a lot of duplication between the app and DB tests. Do we want to change that?

I don't have a strong opinion about this, but in the past we have tried to share duplicate tests [here](https://github.com/aptible/ember-cli-aptible-shared/blob/93caacb80c1135bdd0beb3a32679a30e65963691/test-support/helpers/shared-tests.js)

> This makes requests to the metrics service via the controller. I think I remember seeing somewhere that we wanted to make requests in routes, but I'm not sure. Let me know if I'm imagining things here!

You are correct.  In general, we try to keep AJAX requests in the route.  However, I think this use case is a bit outside of the norm, so I don't think that is a strict requirement.  In lieu of that, though, I do think it'd be good to slim up the controller mixin by moving much of the logic to either a service or a util.  

This is really impressive!

gib:
WOW! Gonna borrow this...

fancyremarker:
This is awesome!!! Very minor style nitpick: I'd vote for truncating the container IDs to 12 characters in the UI, for readability and to match what they'll see in their logs.

krallin:
@fancyremarker Good idea (I did that change in https://github.com/aptible/metrictunnel). Made a few other tweaks (fixed the service type and added command). Here's what this looks like now:

image

krallin:
One new addition: what should the endpoint be for the metrics service? This is hosted via AWS' API gateway, so it essentially has to be 1 domain for prod / staging / sandboxes with a different path.

Right now, the pattern looks like this:

```
https://ewno8ssw8g.execute-api.us-east-1.amazonaws.com/${subdomainSuffix}${environment}/proxy
```

So, for production, the path would be `/production/proxy`. For staging it's `/staging/proxy`. 

Now, for a sandbox with a subdomain suffix (e.g. like in `api-thomas...`), it's: `thomassandbox`.

Questions:

- Do we mind having a amazonaws URL in here despite the fact that this isn't visible to users? I'm leaning towards yes, and we can solve this in API gateway (e.g. use `metrics.aptible.com`).
- Do we prefer `thomassandbox` or `thomas`? In the latter case, we'd drop the bit of information that "this is a sandbox with a suffix".

Thoughts? @fancyremarker @aaw @blakepettersson ?

fancyremarker:
> Do we mind having a amazonaws URL in here despite the fact that this isn't visible to users? I'm leaning towards yes, and we can solve this in API gateway (e.g. use metrics.aptible.com)

Is this just uploading a certificate once? If so that seems relatively easy.

krallin:
Updated the endpoint URIs to something nicer-looking:

- Production:       metrics.aptible.com
- Staging:          metrics.aptible-staging.com
- Sandbox:          metrics.aptible-sandbox/sandbox
- Personal sandbox: metrics.aptible-sandbox/${USER}sandbox

krallin:
@sandersonet Made the tweaks you suggested regarding using a component & a service!

Note: I'll probably update this soon to turn the component into a mixin and have two components (one for memory and one for CPU).

krallin:
Now with CPU metrics:

image

The design may not be ideal.. @sandersonet, @gib, if you have ideas, I'm all ears :)

gib:
Oh man -- this fits right in! Looking great @krallin!!

What does the "close" button close? And what happens when you check the compare to memory limit?

krallin:
@gib:

- Close just takes you back to the services list (it should only show for apps not databases — I'm actually fixing this now).
- Compare to memory limit resizes the chart to show your memory limit (if there is one) no matter what the current usage is:

image

krallin:
(Dammit, I should stop posting test DB passwords in github issues)

gib:
These charts are awesome and placement of actions is fuzzy, I'm reaching to offer feedback here. So with those caveats... I might suggest putting the checkbox on top, closer to the "Memory" title (underneath?) or on the same horizontal line right-aligned or floated right. This would offer a stronger visual grouping of that checkbox action with the Memory chart.

The "close" button is a new navigation pattern. Labeling could offer more clarity... "Back to [Apps | Databases]"?

Neither of these are blockers in my opinion.

sandersonet:
I don't think there are any issues that would be blockers to merging this other than, as Gib mentioned, renaming Close to something like "<- All Services" or something similar.  

I do think it'd be worth circling back to look at the overall UX of services holistically, but that can be saved for another time.  I can open a ticket and assign to myself to do this once SPD is out the door.

sandersonet:
I created https://github.com/aptible/dashboard.aptible.com/issues/556 to track the design task I mentioned above.

krallin:
Thanks @sandersonet!

krallin:
Now using LA instead of CPU, and with a little tooltip explaining what's going on:

image

image


@fancyremarker @aaw @blakepettersson, any feedback on the description of the metrics?
@sandersonet @gib: I'm using `<br><br>` for linebreaks in the tooltip. That may not be ideal..? Let me know what you think!

krallin:
I'm on the fence regarding mentioning that they should install an application performance monitoring tool to go more in depth. Thoughts?

fancyremarker:
The load average explanation looks great. I have two critiques of the memory explanation:

> An upwards trend can be indicative of an issue, but make sure you understand how your app manages memory before jumping to conclusions.

The tone of this sentence is a bit harsh – can we simply omit it? I doubt it would prevent any misguided conclusions anyway, without giving more specific ways to help customers understand their memory usage.

> Note that in the future, memory limits will be strictly enforced on Aptible, regardless of contention on the host.

This sentence is a bit dense, especially since it requires some understanding of how containers are deployed many-to-one on hosts on Aptible. I might just merge into the preceding paragraph, with something like: "[...] apps that exceed their memory limit may be terminated when the underlying host is low on free memory. In the future, apps may be terminated immediately when they exceed their memory limit."

As for mentioning that they should install an application performance monitoring tool, I think it could make sense to link to a support topic about monitoring, where we can discuss the subject in greater depth and recommend some third-party services/tools.

krallin:
@fancyremarker:

> > An upwards trend can be indicative of an issue, but make sure you understand how your app manages memory before jumping to conclusions.
>
> The tone of this sentence is a bit harsh – can we simply omit it? I doubt it would prevent any misguided conclusions anyway, without giving more specific ways to help customers understand their memory usage.

Sure, we could omit it, and perhaps include a more in-depth explanation in a support topic.

> > Note that in the future, memory limits will be strictly enforced on Aptible, regardless of contention on the host.
>
> This sentence is a bit dense, especially since it requires some understanding of how containers are deployed many-to-one on hosts on Aptible. I might just merge into the preceding paragraph, with something like: "[...] apps that exceed their memory limit may be terminated when the underlying host is low on free memory. In the future, apps may be terminated immediately when they exceed their memory limit."

I'm fine with making this change, but does this re-wording entirely solve your concern? It still references "the underlying host" just the same.

> As for mentioning that they should install an application performance monitoring tool, I think it could make sense to link to a support topic about monitoring, where we can discuss the subject in greater depth and recommend some third-party services/tools.

I don't think we can include a link in the tooltip directly (@gib @sandersonet : can we?), but I'll try to find another solution.

krallin:
@fancyremarker: tweaked the messages as you suggested in https://github.com/aptible/dashboard.aptible.com/pull/553/commits/2bfad18262d314039bc2af807ac0eb2c88eb9878

sandersonet:
@krallin  You can include a link, you just need to set the popover's trigger to `click` rather then `hover`.  The downsize to using click is that it then needs to be `clicked` again in order to close the popover.  I think that is fine for now.  Someday we can circle back and add a new component that better handles embedded links in a popover.

fancyremarker:
@krallin:

> I'm fine with making this change, but does this re-wording entirely solve your concern? It still references "the underlying host" just the same.

You're completely right. Should we simply stick to our long-term platform "contract"? In other words, "[...] apps that exceed their memory limit may be terminated when they exceed their memory limit."

krallin:
@fancyremarker: sounds good! Here's where we are now:

image

image

@sandersonet: I tried that after you suggested it (had to tweak the component in ember-cli-aptible-shared directly to do so), but the UX was a little weird / unintuitive, so I'll probably keep this as-is for now. Thanks, though!

krallin:
Also added a little padding to make the LA chart readable when LA is 0:

image

krallin:
@fancyremarker Do we feel it's useful to show the underlying instance ID? (I was talking about this with @chasballew a little earlier)

image

fancyremarker:
@krallin: 

> Do we feel it's useful to show the underlying instance ID? 

I think it would be more useful for us, when impersonating customers, than it would be for customers themselves. I think it would be confusing only for those customers who don't immediately realize it's an EC2 instance ID, but I doubt that would create much overall confusion.

krallin:
@fancyremarker Thanks! I was thinking about this in the context of https://github.com/aptible/sweetness/pull/562 since we'd start showing them there.

I don't really mind either way, this was mostly an experiment to see how it'd look like :)

fancyremarker:
This looks :100: :shipit: to me.

PR #543, opened 2016-03-09
Drag-n-drop .crt, bundled .crt, and .key files to create certificates

gib:
Closes https://github.com/aptible/dashboard.aptible.com/issues/532

I did a bit of hallway user testing (thanks @shahkader & @wcpines ) in hopes this is an improved user experience. You can drop multiple files on the textarea and that makes it tricky to order bundled cert chains, but with https://github.com/aptible/api.aptible.com/pull/327, the ordering will be taken care of on the server and the user will not need to worry about it.

Currently we don’t notify users of ordering problems and they end up seeing them in a browser
(https://support.aptible.com/topics/paas/how-to-order-certs/).

We may be better served by a traditional file upload button. Will watch for feedback.

More thanks: feedback / input from @fancyremarker and ember help from @sandersonet

drag-n-drop-certs

PR #555, opened 2016-03-21
Add new hotness from an ember-cli-aptible-shared update

gib:
* Adds the frame buster (see https://github.com/aptible/dashboard.aptible.com/pull/552)
* Adds the `file-textarea` component and updates labels to improve the UX of adding certs (see https://github.com/aptible/dashboard.aptible.com/pull/543)


tests/acceptance/apps/vhost-new-test.js
```diff
@@ -287,10 +287,6 @@ test(`visit ${appVhostsNewUrl} and create vhost with existing certificates`, fun
 
   signInAndVisit(appVhostsNewUrl);
 
-  stubRequest('post', '/accounts/stubbed-stack/certificates', function() {
-    assert.ok(false, 'should not create a new certificate');
-  });
-
```
sandersonet: Was this causing failures?  With this test case, we shouldn't be creating a new certificate.

gib: I did have failures. I put a break point in all the stubbed requests and that one never got hit. The assert(5) added up to 6 with it, so I removed it, but wasn't sure why it would have caused the test to fail.

gib: Definitely could have been on my end. Maybe Travis will set it straight. I'll read it again.

gib: I'm confused. This should not be passing. I have an update in the works.

gib:
I closed and re-opened to get Travis to run the latest commits. Not sure why it wouldn't update. The red X on a prior commit is from me restarting a successful one and canceling it because I was hoping it would run the latest.

PR #560, opened 2016-03-29
Do not set service container count on scale

krallin:
Found this while I was working through some mismatches between our old permissions model and our new permissions model (see: https://github.com/aptible/auth.aptible.com/pull/203) as reported by Scientist.

It looks like Dashboard currently sets the `container_count` property on a service when scaling it (before it creates the operation. However, I'm not sure that this is expected:

- according to @fancyremarker, users should not be allowed to touch *any* attributes of services
- it definitely looks like Sweetness expects the container count to be the original one when running a service scale operation: https://github.com/aptible/sweetness/blob/5012f50999a108f1ccec4865c45af93422ea3267/lib/sweetness/jobs/multi_instance_scale_service.rb#L7

Am I missing anything? If not, is this the right way to go about fixing this?

--

### DIFF
--- a/app/app/services/route.js
+++ b/app/app/services/route.js
@@ -21,16 +21,12 @@ export default Ember.Route.extend({
 
   actions: {
     scaleService: function(service, containerCount, deferred){
-      service.set('containerCount', containerCount);
-      service.save().then( () => {
-        let op = this.store.createRecord('operation', {
-          type: 'scale',
-          containerCount: containerCount,
-          service: service
-        });
-        return op.save();
-      }).then(deferred.resolve, deferred.reject);
+      let op = this.store.createRecord('operation', {
+        type: 'scale',
+        containerCount: containerCount,
+        service: service
+      });
+      op.save().then(deferred.resolve, deferred.reject);
     }
   }
-
 });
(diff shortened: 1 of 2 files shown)


krallin:
Note: I'm pretty sure this needs to be changed to roll out privileged tokens, otherwise users will get a permission error when they attempt to update the service.

blakepettersson:
FWIW, with #534 that wouldn't be an [issue](https://github.com/blakepettersson/dashboard.aptible.com/blob/add-service-scaler-for-container-size/app/app/services/route.js#L24). If necessary we can split that part out into a separate PR.

krallin:
Thanks @blakepettersson! I'm fine with either merging this PR after it's been reviewed, or splitting that change out from your PR. What's your preference?

(FWIW, this PR is pretty simple, but I'd rather not create more merge conflicts for you to resolve on yours if you'd rather avoid it..!)

blakepettersson:
Hey @krallin, I'm totally fine with splitting the change out of #534! :smile:

sandersonet:
@krallin Thanks for bringing this up.  @aaw brought this up a while ago, but we never addressed it on dashboard: https://github.com/aptible/sweetness/pull/408

blakepettersson:
This has been fixed with #561.

PR #561, opened 2016-03-29
Reload operation on service scale

blakepettersson:
This commit is extracted from #534; this reloads the operation until
either it times out, fails or succeeds.

cc @krallin


krallin:
Thanks @blakepettersson !

PR #566, opened 2016-04-01
Add RSS vs + cache/buffers metrics

krallin:
This updates the container memory graph to show RSS (which is memory that usually cannot be reclaimed and *will* result in the container being terminated if it exceeds its allocation) by default instead of RSS + caches / buffers:

image

This also adds a toggle to let the user visualize RSS vs. RSS + caches / buffers:

image

This can get a little crowded with a lot of containers:

image

But it's of course possible to hide some containers to drill down:

image

I also updated the guidance to give a few explanations about those metrics:

image

(This is dependent on deploying https://github.com/aptible/metrictunnel/pull/8 to expose the `memory_rss` and `memory_all` metrics).

cc @fancyremarker: let me know if this is what you had in mind when we discussed yesterday.
cc @sandersonet the last commit has a message that explains the issue I was running into (and why I'm trying to mock the `{{c3-chart}}` component). Looks like the CI tests might fail with  the same reason too.

### DIFF
--- a/app/components/c3-chart/component.js
+++ b/app/components/c3-chart/component.js
@@ -70,12 +70,13 @@ export default Ember.Component.extend({
 
   dataDidChange: Ember.observer('data', function() {
     ensurePromise(this.get('data')).then((data) => {
-      if (Ember.isEmpty(data)) {
+      if (Ember.isEmpty(data) || Ember.keys(data).length === 0) {
         return;
       }
 
       let chart = this.get('_chart');
       if (chart) {
+        data.unload = true;
         chart.load(data);
       } else {
         this.get('_submitInitialData')(data);
--- a/app/components/c3-chart/component.js
+++ b/app/components/c3-chart/component.js
@@ -8,6 +8,18 @@ function ensurePromise(x) {
     });
 }
 
+function withObservedChartAttribute(attrName, handler) {
+  return Ember.observer(attrName, function () {
+    ensurePromise(this.get(attrName)).then((attrValue) => {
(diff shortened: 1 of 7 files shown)


krallin:
(I'll work on fixing the test failure thanks to @sandersonet's link)

fancyremarker:
This looks excellent, and the copy is perfect! Feel free to :shipit: when all dependencies are in place.

krallin:
I finally tracked down the failing tests to the c3chart object being torn down before all transitions had finished, which caused it to access fields that no longer existed. Unfortunately, solving the issue isn't really possible (I included a large number of comments in my last commit explaining why), so instead I'm working around it by not tearing it down immediately (and instead waiting for 500ms so transitions can finish). 

cc @sandersonet, if you'd like to check that i'm not doing something completely crazy / stupid? : )

app/components/c3-chart/component.js
```diff
@@ -61,60 +73,84 @@ export default Ember.Component.extend({
...
+     * one did.
+     *
+     * Now, we could use the same callbacks to ensure we don't schedule a
+     * transition if we already scheduled one, but at this point we're adding a
+     * substantial amount of complexity, and we run the risk of locking up the
+     * chart if for some reason the onrendered callback isn't called (which
+     * seems possible considering c3.js makes no guarantees as to when this
+     * callback will be called), which doesn't seem worth if the only upside is
+     * saving a few seconds on running tests.
+     */
```
gib: Thanks for the clear comments

PR #567, opened 2016-04-05
Update LA description to account for blocked tasks

krallin:
After rolling out https://github.com/aptible/cadvisor-factory/pull/7, we'll want to update our copy to match reality.

cc @fancyremarker


fancyremarker:
👍

PR #570, opened 2016-04-11
Add disk metrics

krallin:
This adds support for disk metrics. The metrics are included for DBs at the bottom of the metrics tab, and look like this:

image

A few notes here:

- First: this isn't ready for merging yet (I need to make the PRs in https://github.com/aptible/metrictunnel and https://github.com/aptible/cadvisor-factory)
- Second: the numbers reported aren't *exactly* the same as the numbers requested (i.e. someone asks for a 10GB volume, this reports 9.7GB). AFAICT, the numbers reported here are actually correct — the space that's "missing" being space that's used for accounting by the filesystem (and perhaps a little space reserved for root as well). 

### DIFF
--- a/app/components/container-disk-metrics/component.js
+++ b/app/components/container-disk-metrics/component.js
@@ -1,0 +1,9 @@
+import Ember from 'ember';
+import ContainerMetricsComponentMixin from "diesel/mixins/components/container-metrics";
+
+export default Ember.Component.extend(ContainerMetricsComponentMixin, {
+  metric: "fs",
+  axisLabel: "Disk usage",
+  axisFormatter: (v) => `${v} GB`,
+  axisBottomPadding: 0  // Disk usage is always > 0, no need for extra padding
+});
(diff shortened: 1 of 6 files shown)


krallin:
Small tweak: the device should actually be shown on that graph:

image

gib:
So awesome!

I wonder if there is a way to simplify the 10GB / 9.72GB discrepancy by reporting percentage used.

sandersonet:
:+1:

krallin:
@gib I'm a little concerned showing the `%` would hide some valuable information, so if possible I'd rather not go that route. I reckon we could explain the discrepancy in the helper text for this metric; though!

gib:
Just thinking through options... maybe start the empty db at 0.28GB (with helper text) or labeling the max as 9.72 instead of 10 with helper text (in order to eliminate the limit line)?
I definitely see what you're saying... could cause unnecessary problems for customers trying to debug something ("Where did this extra 0.28GB came from?!?").

chasballew:
+1 for explaining in the helper text

krallin:
Proposed helper text:

image

fancyremarker:
👍  on helper text.

PR #573, opened 2016-04-15
Fix domain type label for new default vhosts

gib:
The form' domain type label still had the old `aptible.in` domain. Updated:

screen shot 2016-04-14 at 11 32 34 pm


gib:
Leeeeeeroy Jenkins

PR #575, opened 2016-04-19
Disable memory slider on v1 stacks

blakepettersson:
Fix typo where memory slider will always be enabled on a v1 stack.

### DIFF
--- a/app/components/service-scaler/component.js
+++ b/app/components/service-scaler/component.js
@@ -8,14 +8,12 @@ export default Ember.Component.extend({
 
   isSliding: false,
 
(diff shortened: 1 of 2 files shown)


sandersonet:
Awesome, thanks for the added test too.

You could simplify this to just:

```javascript
isv2Stack: Ember.computed.equal('service.stack.sweetnessStackVersion', 'v2'),
isv1Stack: Ember.computed.not('isv2Stack'),
```

Your call if you want to make the change. 

Thanks!

blakepettersson:
> You could simplify this to just:

Fixed and rebased.

PR #576, opened 2016-04-20
Adjust RAM slider range to 512-8192 MB

fancyremarker:
Opening this PR to increase the minimum RAM size from 256 MB to 512 MB, per [a Slack discussion](https://aptible.slack.com/archives/ops/p1461096869000006).

### DIFF
--- a/app/components/service-scaler/template.hbs
+++ b/app/components/service-scaler/template.hbs
@@ -61,13 +61,13 @@
           {{no-ui-slider didSlide="setContainerSize"
                          didSet="finishSliding"
                          start=containerSize
-                         rangeMin=256
+                         rangeMin=512
                          rangeMax=8192
                          snap=true
                          disabled=shouldDisable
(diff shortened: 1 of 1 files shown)


app/components/service-scaler/template.hbs
```diff
@@ -61,13 +61,13 @@
           {{no-ui-slider didSlide="setContainerSize"
                          didSet="finishSliding"
                          start=containerSize
-                         rangeMin=256
+                         rangeMin=512
                          rangeMax=8192
                          snap=true
                          disabled=shouldDisable
-                         range="256,512,1024,2048,4096"}}
+                         range="512,1024,2048,4096"}}
```
sandersonet: Missed an `8192` here.

sandersonet: NM looks like it doesn't need to be specified in the range.

sandersonet:
:+1:

PR #580, opened 2016-05-03
Add container size in billing overview

blakepettersson:
This is still a bit rough around the edges, and the tests needs to be updated but I would like some feedback on the ongoing work to add custom container sizes in the billing overview. The current result can be seen below, and this itemises all entries with the price so that it's more clear for a customer.

cc @gib @fancyremarker @sandersonet 

screen shot 2016-05-03 at 17 57 37

### DIFF
--- a/app/components/disk-resource-usage/component.js
+++ b/app/components/disk-resource-usage/component.js
@@ -1,0 +1,13 @@
+import Ember from 'ember';
+
+export const HOURS_PER_MONTH = 730;
+
+export default Ember.Component.extend({
+  tagName: 'tr',
+
+  hasDisk: Ember.computed.notEmpty('database.disk'),
+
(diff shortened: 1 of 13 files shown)


wcpines:
@blakepettersson any chance we could provide environment information as well? 

(Similar to and may obviate the need for https://github.com/aptible/dangerzone/issues/65)

blakepettersson:
@wcpines it's not very visible, but there is already environment information on the billing overview. The `sdfsdf` (good name, I know) on the picture is actually the environment name. All of the resource views iterate over each environment and prints its name along with all its resources under it.

sandersonet:
👏 Excellent

blakepettersson:
@sandersonet the tests are failing with an obscure error on travis, but they pass locally. Do you have any idea what that could be?

gib:
This is awesome!

PR #587, opened 2016-05-09
UI tweaks to #580

blakepettersson:
This is a PR that adjusts some issues with #580. In some cases the promises were never fully resolved in the UI, which led to some calculations being incorrect. I refactored the promise resolving so that it all happens within the route instead of trying to rely on computed properties to resolve the promises.

The rest is pretty much simply refactoring so that the affected components are easier to test, as well as changing the rendering of vhost rows to only render if its `status` is `provisioned`, and only rendering disk rows if the disk has a non empty `size`.

cc @sandersonet @gib

### DIFF
--- a/app/components/disk-resource-usage/component.js
+++ b/app/components/disk-resource-usage/component.js
@@ -5,7 +5,7 @@ export const HOURS_PER_MONTH = 730;
 export default Ember.Component.extend({
   tagName: 'tr',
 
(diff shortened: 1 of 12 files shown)


sandersonet:
Great refactor.  Thanks!

gib:
Ditto! Looks awesome!

PR #584, opened 2016-05-06
WIP - Add support for 2FA

krallin:
View https://github.com/aptible/auth.aptible.com/pull/208 for more information. This isn't fully ready yet (see the WIP list below), but I'd love to get some early feedback :).

This:

- Allows users to enable 2FA.
- Allows users to login using 2FA.
- Allows users to provide a 2FA token when updating their email or
  password.

WIP:

- Warn users that 2FA isn't enforced on git pushes.
- Prompt users to update their aptible CLI.
- Prompt users to save recovery tokens.

cc @fancyremarker @sandersonet @gib

PR #588, opened 2016-05-11
Support 2fa elevation

krallin:
This replaces #584 and works with https://github.com/aptible/auth.aptible.com/pull/209 (it also requires an ember-cli-aptible-shared update).

All the major commits have commit messages, so I'll refer you to those rather than repeat them here 😄 

cc @sandersonet @gib @fancyremarker


sandersonet:
👍 Amazing work @krallin !! 👍

krallin:
Added 2 extra commits to address a few UX quirks that came up while user testing with @chasballew and @wcpines.

@sandersonet, if you'd like to give those an look, I'd appreciate it! 

Thanks!

wcpines:
Very minor, but I'm curious why the 1pass QR reader failed.  Any thoughts?

krallin:
I haven't been able to find out why that might be the case.

I did some research this morning, which essentially boiled down to finding
a post where 1password folks acknowledge it's pretty badly broken...

That being said, you can now reveal the underlying OTP secret, which does
work if you set it as the value for a field of type one-time password in
1password (which I think is good enough)! That's actually how I smoke
tested it :).

Cheers,

wcpines:
Thanks for digging into that a bit!  And yes, I was also able to get the otpauth URL to work when I was testing with you. Glad to confirm there's a viable work-around :)

PR #589, opened 2016-05-11
See provision/deprovision failures in the UI (WIP)

blakepettersson:
This PR adds deprovisioning and deprovisioning failures to the UI, that looks something like the picture below for vhosts, log drains and databases, as well as deprovisioning apps.

There are still outstanding issues - for some reason I cannot get the log drains to consistently update in the UI even though the status has changed in the API, as well as having to fix the failing tests. 

screen shot 2016-05-10 at 16 44 37


blakepettersson:
The outstanding issues have been fixed afaik, do you have any further comments @sandersonet or @gib?

gib:
Fixed here in shared I think: https://github.com/aptible/ember-cli-aptible-shared/pull/152

app/components/list-databases/component.js
```diff
@@ -0,0 +1,46 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  sortBy: ['handle:asc'],
+
+  arrayChanged: function() {
+    Ember.run.once(this, 'checkModelArray');
+  }.observes('model.[]'),
```
gib: It's my understanding that the `.observes` syntax has been replaced by `Ember.computed`, but should be the same. This is bound to changes on the length of the model, so it should be called every time anything is added or removed from `model`, which makes calling `checkModalArray` redundant, along with sending the action `redirect` first... why not send `forceRedirect` all in one go?

```javascript
arrayChanged: Ember.computed('model.[]', function() {
  if (this.get('model.length') === 0) {
    this.sendAction('forceRedirect');
  }
}),
```

gib:
I support merging as is. I added a comment on how I understand things to work that could perhaps reduce redundancy, but maybe you found the need for more methods in working with the code?

I'm happy to kick the tires locally and on staging too after the meeting this morning.

app/components/list-databases/component.js
blakepettersson: The reason there's an `Ember.run.once` is that it will otherwise call `checkModelArray` for each element in the model, at least when I tested this.

PR #599, opened 2016-05-18
Add tests for activating-item

krallin:
This adds a test that would have caught the bug in https://github.com/aptible/ember-cli-aptible-shared/pull/123

cc @gib @sandersonet


krallin:
Note: this test is failing, but that's the point :)

gib:
red-green! Thank you @krallin

Issue #598, opened 2016-05-17
Misc UI improvements


chasballew:
In the environment header and app index, it would be helpful to see RAM footprints and totals:

<img width="498" alt="app_index" src="https://cloud.githubusercontent.com/assets/3422584/15330673/1bfc654a-1c1b-11e6-962e-2028a2cf5eee.png">

Also in the environment header, some log drains appear to be missing:

![missing_log_drains](https://cloud.githubusercontent.com/assets/3422584/15330691/2e3a6aea-1c1b-11e6-9947-569c801f7df7.png)

Again, in the database index, it would be helpful to see the RAM footprint of a database, in addition to disk size:

<img width="371" alt="database_index" src="https://cloud.githubusercontent.com/assets/3422584/15330713/44129c0c-1c1b-11e6-92b3-f7af59d700c3.png">

Finally, 1) it seems more helpful to display the "Metrics" tab as the default tab for a database, 2) the "Size" header should include both RAM and disk, and should not break over a line, and 3) the connection URL should not break over a line:

<img width="511" alt="database_view" src="https://cloud.githubusercontent.com/assets/3422584/15330753/74e6367c-1c1b-11e6-8e6f-6e5510ba2264.png">

@gib

chasballew:
Additions:

- Under each app service, it might be helpful to see associated endpoints inline
- A link to delete a certificate would be helpful, at the bottom of the endpoint creation view:

<img width="289" alt="screenshot 2016-06-28 13 05 34" src="https://cloud.githubusercontent.com/assets/3422584/16428949/8ecdd626-3d31-11e6-8249-c67870a10aae.png">

chasballew:
#634 probably also belongs in this burndown

chasballew:
- When trying to delete a certificate, customers would benefit from clearer messaging indicating how to  update or delete the endpoints that the deletion action

PR #600, opened 2016-05-18
Fix log drain count header

krallin:
We show host:port for one log drain in the stack header, but if the log
drain is a log tail, that breaks down (because it does not have those).

This updates the logic to prefer showing a log drain that's not a log
tail, and shows nothing if the log drain is a log tail.

---

This fixes one of the issues reported in https://github.com/aptible/dashboard.aptible.com/issues/598

### DIFF
--- a/app/components/stack-metadata/component.js
+++ b/app/components/stack-metadata/component.js
@@ -24,5 +24,11 @@ export default Ember.Component.extend({
   vhostNamesTooltip: Ember.computed('model.vhostNames', function() {
     let names = this.model.get('vhostNames');
     return names.slice(this.get('maxVisibleDomainNames')).join(', ');
-  })
+  }),
+
+  logDrainsForSnippet: Ember.computed.filter("model.logDrains", (o) => {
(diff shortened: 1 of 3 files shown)


app/components/stack-metadata/component.js
```diff
@@ -24,5 +24,11 @@ export default Ember.Component.extend({
   vhostNamesTooltip: Ember.computed('model.vhostNames', function() {
     let names = this.model.get('vhostNames');
     return names.slice(this.get('maxVisibleDomainNames')).join(', ');
-  })
+  }),
+
+  logDrainsForSnippet: Ember.computed.filter("model.logDrains", (o) => {
+    return !(o.get("isLogTail"));
+  }),
```
krallin: Note: I tried using `filterBy` here, but for some (obscure)  reason, it wouldn't work. So.. This is what we get instead.

gib:
sweet!

sandersonet:
Thanks for this :+1:

PR #602, opened 2016-05-20
Show 2FA enrollment status

krallin:
We probably want to display whether users have enabled 2FA on the users page. I don't have a very good design for it, so perhaps @gib and @sandersonet have better ideas, but here's at least a starting point:

image

If this works, I'll also add tests for the component / page. Just let me know.

cc @fancyremarker @chasballew

### DIFF
--- a/app/components/otp-status/component.js
+++ b/app/components/otp-status/component.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+
+const makeComputedToggle = function(enabledValue, disabledValue) {
+  return Ember.computed('otpEnabled', function() {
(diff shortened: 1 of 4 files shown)


sandersonet:
For now I think we ship with this, but eventually we'll want to revisit this screen and probably change the layout to tables/rows.

A potential redesign of the members index is included here: 
[members-and-roles.pdf](https://github.com/aptible/dashboard.aptible.com/files/278474/members-and-roles.pdf)

krallin:
Sounds good; thanks @sandersonet. Added tests and merging (once CI finishes).

PR #603, opened 2016-05-20
Fix "rebuilded" label

krallin:
cc @gib @sandersonet


sandersonet:
you mean rebuilded isn't a word? 👍

PR #604, opened 2016-05-20
Cleanup impersonation

krallin:
This cleans up impersonation by replacing the convoluted "logout and then try to log back in as the use and if you fail try logging back in as the admin" workflow for impersonation with a much more streamlined one, which first acquires the user token (and persists it to cookies), and upon success destroys the admin token and reloads the UI .

One side benefit is that we never make *any* request with the user's token before we destroy the admin token, which is arguably a little safer.

cc @sandersonet @gib

### DIFF
--- a/app/logout/route.js
+++ b/app/logout/route.js
@@ -8,7 +8,7 @@ export default Ember.Route.extend({
     }
   },
   activate: function(){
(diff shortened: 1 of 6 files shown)


sandersonet:
Love the refactor, thanks! :+1:

PR #610, opened 2016-05-25
Fix more activating tabs

krallin:
Follow up to ember-cli-aptible-shared#123, which fixed the
`activating-item` component, but that component isn't being used here.

cc @gib @sandersonet


sandersonet:
👍

gib:
This is a nice change

PR #620, opened 2016-06-06
Set bs-html when HTML interpolation is required

krallin:
See https://github.com/aptible/ember-cli-aptible-shared/pull/133

Will update this PR with the new ember-cli-aptible-shared version once it's released, so we can ensure the tests run against that new lib. 

---

cc @sandersonet @gib


sandersonet:
👍 Perfect

PR #622, opened 2016-06-12
Use "endpoints" instead of "domains" or "Vhosts"

chasballew:
The intent is to update only the user-facing components, leaving the internal code for now.

This terminology change is also reflected in aptible/www.aptible.com#218 and aptible/support#268

@sandersonet - We can discuss, but I wasn't able to get a local environment running properly to test the UI. There is still one test failure:

```
not ok 72 PhantomJS 2.1 - Acceptance: App Endpoint New: visit /apps/1/vhosts/new should remove certificate form if default endpoint is clicked
    ---
        actual: >
            false
        expected: >
            true
        stack: >
            http://localhost:7357/assets/test-support.js:3830:18
            exception@http://localhost:7357/assets/vendor.js:53385:9
            onerrorDefault@http://localhost:7357/assets/vendor.js:44082:33
            trigger@http://localhost:7357/assets/vendor.js:68351:19
            _onerror@http://localhost:7357/assets/vendor.js:69317:29
            publishRejection@http://localhost:7357/assets/vendor.js:67624:23
            http://localhost:7357/assets/vendor.js:44051:15
            invoke@http://localhost:7357/assets/vendor.js:11571:20
            flush@http://localhost:7357/assets/vendor.js:11635:17
            flush@http://localhost:7357/assets/vendor.js:11436:22
            end@http://localhost:7357/assets/vendor.js:10725:30
            run@http://localhost:7357/assets/vendor.js:10847:21
            run@http://localhost:7357/assets/vendor.js:31688:32
            http://localhost:7357/assets/vendor.js:53624:39
        message: >
            Error: Element [object Object] not found.
        Log: |
    ...
```

### DIFF
--- a/app/components/usage-quote-by-resource/component.js
+++ b/app/components/usage-quote-by-resource/component.js
@@ -2,13 +2,23 @@ import Ember from 'ember';
 
 export const HOURS_PER_MONTH = 730;
 
+const RESOURCE_LABELS = {
+  'domain': 'endpoint',
+  'disk': 'disk',
+  'container': 'container'
+};
+
 export default Ember.Component.extend({
-  resourceLabel: Ember.computed('resource', function() {
-    if(this.get('resource') === 'domain') {
-      return 'Endpoints';
-    } else {
-      return this.get('resource').capitalize().pluralize();
-    }
+  resourceHandle: Ember.computed('resource', function() {
+    return this.get('resource').toLowerCase();
+  }),
+
+  resourceTitle: Ember.computed('resourceHandle', function() {
+    return RESOURCE_LABELS[this.get('resourceHandle')].capitalize().pluralize();
+  }),
+
+  resourceLabel: Ember.computed('resourceHandle', function() {
+    return RESOURCE_LABELS[this.get('resourceHandle')];
   }),
 
   diskUsage: Ember.computed.mapBy('stacks', 'totalDiskSize'),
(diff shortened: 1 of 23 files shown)


app/components/usage-quote-by-resource/component.js
```diff
@@ -4,7 +4,11 @@ export const HOURS_PER_MONTH = 730;
 
 export default Ember.Component.extend({
   resourceLabel: Ember.computed('resource', function() {
-    return this.get('resource').capitalize().pluralize();
+    if(this.get('resource') === 'domain') {
```
chasballew: @sandersonet - Gross

chasballew:
Passing in the resource (https://github.com/aptible/dashboard.aptible.com/blob/master/app/components/usage-quote-by-resource/template.hbs#L9) means that I think we'd need to change some of the JS to get the billing view to work:

screenshot 2016-06-12 15 28 23

sandersonet:
@chasballew I can fix the broken test if you give me contributor access to your fork.

sandersonet:
@chasballew - I fixed the broken test and updated the domains label on billing overview.  Thank you very much for this PR!

This was a surprise to me, but apparently the `click()` helper throws that ugly error if you pass an empty  jQuery collection.  The fix (and best practice probably) here would be to always use `findWithAssert` rather than just `find`--especially when you intend to interact with the element.    In this case, the `find()` was not returning any results, so `findWithAssert` would have failed with a more meaningful error.

PR #625, opened 2016-06-13
Add UI for creating and browsing database backups

krallin:
This adds a new UI tab for managing backups (i.e. viewing existing backups and scheduling a new manual backup).

image

This does not (yet) allow for restoring a backup via the UI, mainly because the resulting UX would be confusing, because there is nothing to redirect to after you restore a backup (the new database doesn't exist yet, and there is no "pending operations" view). 

So, for the time being, I suggest we support this via the CLI.

### DIFF
--- a/app/components/backup-restore-link/component.js
+++ b/app/components/backup-restore-link/component.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  // classNameBindings: [':operation-item'],
+  content: Ember.computed("backup.copiedFrom", "backup.database", function() {
+    // TODO: Link to CLI install instructions?
+    // Restoring from a copy is equivalent to restoring from the original,
+    // except it's slower, so we offer the original's ID if there's one.
+    const restoreId = this.get("backup.copiedFrom.id") || this.get("backup.id");
+    const bits = [];
+    bits.push('<p>Use the Aptible CLI to restore this backup to a new database:</p>');
+    bits.push(`<p><code>aptible backup:restore ${restoreId}</code></p>`);
+    return bits.join(" ");
+  })
+});
--- a/app/components/backup-restore-link/template.hbs
+++ b/app/components/backup-restore-link/template.hbs
@@ -1,6 +1,6 @@
 {{#bs-popover
   bs-trigger="click"
-  bs-container=".ember-view.application-wrapper"
+  bs-container=bindPopover
   bs-html=true
   title="Aptible CLI"
   content=content}}
--- a/app/components/backup-restore-link/template.hbs
+++ b/app/components/backup-restore-link/template.hbs
@@ -1,0 +1,8 @@
+{{#bs-popover
+  bs-trigger="click"
+  bs-container=".ember-view.application-wrapper"
+  bs-html=true
+  title="Aptible CLI"
+  content=content}}
+  <a>Restore to a New Database</a>
+{{/bs-popover}}
--- a/app/database/backups/template.hbs
+++ b/app/database/backups/template.hbs
@@ -1,0 +1,37 @@
+<div class="row">
+  <div class="col-xs-8">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3>Create a New Database Backup</h3>
+      </div>
+      <div class="panel-body">
+        <p>
+        You can initiate the creation of a new backup at any time. Existing
+        backups are shown below. Note that if you schedule a manual backup,
+        Aptible won't schedule an automated backup for another 24 hours.
+        </p>
+        <button class="btn btn-success btn-lg" {{action 'createBackup'}}>
+          Create New Backup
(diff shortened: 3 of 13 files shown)


gib:
This is fantastic! I think the flow of copy/paste a command for the cli is great solution to get this in. This goes a long way to addressing the customer questions / requests around back ups I'm familiar with. 

I think we will replace the floating panels with a generic table style that @sandersonet has been using in the most recent designs, but definitely not a blocker.

krallin:
Thanks @gib! I originally had a list here, but it looked a little awkward, so I went with the panels (this also removed the need for ad-hoc styling). Happy to make tweaks when needed.

sandersonet:
:+1: Perfect

app/database/backups/template.hbs
```diff
@@ -0,0 +1,37 @@
...
+  <div class="col-xs-8">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3>Create a new database backup</h3>
+      </div>
+      <div class="panel-body">
+        <p>
+        You can initiate the creation of a new backup at any time. Existing
+        backups are shown below. Note that if you schedule a manual backup,
+        Aptbile won't schedule an automated backup for another 24 hours.
```
fancyremarker: s/Aptbile/Aptible/

app/components/backup-restore-link/component.js
```diff
@@ -0,0 +1,15 @@
...
+export default Ember.Component.extend({
+  // classNameBindings: [':operation-item'],
+  content: Ember.computed("backup.copiedFrom", "backup.database", function() {
+    // TODO: Link to CLI install instructions?
+    // Restoring from a copy is equivalent to restoring from the original,
+    // except it's slower, so we offer the original's ID if there's one.
+    const restoreId = this.get("backup.copiedFrom.id") || this.get("backup.id");
+    const bits = [];
+    bits.push('<p>Use the Aptible CLI to restore this backup to a new database:</p>');
+    bits.push(`<p><code>aptible backup:restore ${restoreId}</code></p>`);
```
fancyremarker: This is more a question for a pending aptible-cli PR, but will this command expect/allow a handle? Will it default to using `"#{db.handle}-#{backup.id}"` for the database created from backup, so that this command will work exactly as displayed?

fancyremarker:
👍 , looks awesome. Panel display would not have been my initial thought, but I think it looks great!

app/components/backup-restore-link/component.js
krallin: @fancyremarker My plan is:

- Default to the handle + restore timestamp (rather than ID — I think it's more expressive).
- Configurable via `--handle`.

krallin:
@fancyremarker I think it makes the backups feel a little more tangible! 😄 

(I also suck at CSS and as noted above it makes me sidestep CSS)

app/database/backups/template.hbs
krallin: Whoops; thanks!

app/database/backups/template.hbs
```diff
@@ -0,0 +1,37 @@
...
+        <h3>Create a new database backup</h3>
+      </div>
+      <div class="panel-body">
+        <p>
+        You can initiate the creation of a new backup at any time. Existing
+        backups are shown below. Note that if you schedule a manual backup,
+        Aptible won't schedule an automated backup for another 24 hours.
+        </p>
+        <button class="btn btn-success btn-lg" {{action 'createBackup'}}>
+          Create new backup
```
chasballew: Let's use initial caps in button labels for consistency, e.g. "Create New Backup". I can put something in a style guide for this, if it would help.

app/components/backup-restore-link/template.hbs
```diff
@@ -0,0 +1,8 @@
+{{#bs-popover
+  bs-trigger="click"
+  bs-container=".ember-view.application-wrapper"
+  bs-html=true
+  title="Use the Aptible CLI to restore"
```
chasballew: This can be just "Aptible CLI"

app/components/backup-restore-link/template.hbs
```diff
@@ -0,0 +1,8 @@
...
+  content=content}}
+  <a>Restore to a new database</a>
```
chasballew: "Restore to a New Database"

app/database/backups/template.hbs
```diff
@@ -0,0 +1,37 @@
+<div class="row">
+  <div class="col-xs-8">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3>Create a new database backup</h3>
```
chasballew: minor style tweak, "Create a New Database Backup"

krallin:
Thanks for your comments @chasballew , I fixed the style nits

fancyremarker:
Closed and reopened as #631, to incorporate ember-cli-aptible-shared version bump.

PR #626, opened 2016-06-13
Update Org Membership / Pending Invitations Templates

gib:
- data listed in tables
- members and pending invitations are at separate routes
- introduces a new navigation control for viewing the members and pending invitations
- sets up UI for pending auth / role updates

Adding a little more test coverage, but looking for feedback and any blockers to shipping in the routes and controller setup. @sandersonet 

members-invites-ui


app/organization/members/edit/template.hbs
```diff
@@ -1,61 +1,57 @@
...
+                      organizationRole=role
+                      changeset=changeset}}
+                    <label>{{role.name}}</label>
+                  </li>
+                {{/each}}
+              </ul>
             </div>
-          {{/unless}}
+          </form>
         </div>
```
sandersonet: While we're in here, can you expand the form tag to also include the save/cancel buttons?

app/organization/members/index/template.hbs
```diff
@@ -1,53 +1,36 @@
...
-        <div class="row">
-          {{#each sortedInvites as |invitation|}}
-            <div class="col-xs-4">
-              {{partial 'organization/invited-user'}}
-            </div>
-          {{/each}}
-        </div>
-      </div>
+<div class="sort-group organization-members">
+  <ul class='aptable__nav'>
```
sandersonet: Can we reuse the nav pills here for consistency?  I agree that we want to revisit those styles, but I think these tabs are a bit too subtle.  See the [app nav tabs](https://github.com/aptible/dashboard.aptible.com/blob/master/app/templates/app/-nav.hbs) as an example

app/organization/members/index/template.hbs
```diff
@@ -1,53 +1,36 @@
...
-          {{#link-to 'organization.invite' class="btn btn-primary"}}
-            Invite New Member
-          {{/link-to}}
-        {{/if}}
-      {{/aptible-ability}}
-    </div>
-  </div>
-</div>
-
+  <table class="aptable">
```
sandersonet: `aptable` nailed it!

app/styles/components/_aptables.scss
```diff
@@ -0,0 +1,106 @@
+//
+// Aptible Table Component
+//
+.aptable {
+  border: 1px solid #E0E0E0;
+  box-shadow: 0px 2px 4px 0px rgba(0,0,0,0.06);
+  border-radius: 3px;
+  border-collapse: separate;
+  width: 100%;
+
```
sandersonet: It may be worth considering moving these styles to a parent container rather than directly on the table.  Many of our future tables include some metadata within the same container--unless we're going to do a `rowspan`, we'll need these styles on the parent container so we can pair the metadata with a table in the same container. 

e.g. https://www.dropbox.com/s/tv8ml0gq7kfc547/Screenshot%202016-06-13%2011.22.19.png?dl=0

sandersonet:
👍 Fantastic!

app/organization/members/edit/template.hbs
gib: Will do!

app/organization/members/index/template.hbs
gib: Yes. It is pretty subtle. I also tried using boostrap tabs, but it looked a little weird right under the the header border, but header needs the border.

gib: Hey @sandersonet... I was also thinking of adding Pending Invitations to the left nav. What do you think?

sandersonet: I think having it linked in the left nav OR in the header tabs works, just not both.   Generally the header tabs are for subsections of the current left-nav item.  I think removing it from the tabs and adding it to the left works.

app/styles/components/_aptables.scss
gib: Thanks for bringing this up. After thinking about it... I may go for a new class that uses an `aptable` in its composition.

`aptable-group` perhaps? With an `aptable--no-border`. Going to think about it a bit more.

gib:
screen shot 2016-06-15 at 1 24 06 pm

PR #629, opened 2016-06-15
Make coupon usage label like credit section

wcpines:
Add a little context for what you're looking at instead of just repeating the coupon ID


gib:
A little more context, currently looks like: 

This change replaces the lower gray duplicate coupon name with the copy: "Any discounts or coupons applied to your subscription"

Awesome, thanks @wcpines!

PR #635, opened 2016-06-23
Revoke tokens

krallin:
As demoed earlier today,

This adds a "log out all sessions" button on the account page, and
prompts to log out all sessions when the user updates their email or
password, or enables 2FA (not when they disable it, though).

---


sandersonet:
👍

krallin:
@sandersonet @gib @fancyremarker : I've thrown in an additional commit to add a logout page that includes the revoke all tokens functionality, which is probably the best place to put it to ensure that users are familiar with it (for reference, this is similar to what e.g. Stack Overflow does). 

This also has the side benefit of not letting pranksters trick people into logging out by sending them a link to the logout page (which isn't really a big deal though).

The page looks like this:

image

Let me know what you think. If we do decide to roll this out, we'll want to set up an Intercom announcement ahead of time so that users are notified that this is changing when logging in.

Cheers,

PR #640, opened 2016-07-04
Add support for ACME VHOSTs 

krallin:
This is what I demo'ed last week. This lets customers provision Endpoints via ACME. In the Dashboard, those are branded as "Managed SSL".

There are a few changes here:

- First, I've made a bunch of tweaks to the new VHOST UI:
  - Replaced helper popovers on labels with a "?" popover (consistent with e.g. metrics). In practice, I couldn't get the helper popovers to show otherwise (IOW: the current popovers weren't working AFAICT). Accordingly, I showed disabled inputs when a given type is unavailable (we used to not show anything, which would have made the "?" popover confusing). I had to add them manually since our dependency that provides radio buttons doesn't support disabled inputs in the version we have (and upgrading seemed to be a little complex).
  - Added the new type to the input. It's only shown on v2 stacks. It does some very light validation that the input remotely looks like a domain, but is intentionally lax so as to not accidentally refuse domains that don't work.
  - I refactored most of the logic of the create VHOST component. Mainly: moved all data access to the route (this eliminates flicker when the form loads), replaced most implicit conditionals (e.g. if the VHOST is non-default, then it needs a cert) with more explicit ones (i.e. computeds like `certificateNeeded`). These changes were mainly needed to keep the component readable despite the addition of new logic.
- Second, I've added a new (somewhat extensible) "Action Needed" status to VHOSTs. This allows a component to be loaded for each Action Needed status. Right now, the only such status is an ACME VHOST that needs to be initialized (that's what provides the "I Created The CNAME" button).

A few screenshots:

imageimage

image

image

image

image

---

### DIFF
--- a/app/components/create-vhost/template.hbs
+++ b/app/components/create-vhost/template.hbs
@@ -20,51 +20,99 @@
               {{ ultimate-select items=services name="service" itemKey="id" itemValue="handle" update=(action (mut vhostService)) selected=vhostService autofocus=true }}
             </div>
 
-          {{#unless serviceHasDefaultVhost}}
               <div class="form-group">
-                <label class="block" for="domain-type">Endpoint type:</label>
+                <label class="block" for="domain-type">
+                  Endpoint type:
+                    {{more-info-icon
+                    title="Choosing an Endpoint Type"
+                    bs-html=true
+                    body="
+                    <p>
+                    For development apps and non-user facing APIs, choose a
+                    <strong>Default Endpoint</strong>: Aptible takes care of
+                    the domain name and SSL certificate for you.
+                    </p>
+                    <p>
+                    For other use cases, a <strong>custom domain with managed
+                    SSL</strong> is a good option: you use your own domain
+                    name, but Aptible handles provisioning and renewing your SSL
+                    certificate.
+                    </p>
+                    <p>
+                    For more advanced use cases, like using an Extended
+                    Validation SSL certificate, use a <strong>custom domain
+                    with a custom SSL certificate</strong> for complete control.
+                    </p>"}}
+
+                </label>
                 <div class="radio">
+                  {{#if defaultVhostAllowed}}
+                    <div>
+                      {{#radio-button value=TYPE_DEFAULT groupValue=vhostType name="domain-type"}}
+                        Use app-{{ model.app.id }}.on-aptible.com default endpoint.
+                      {{/radio-button}}
+                    </div>
+                  {{/if}}
                   <div>
-                    {{#radio-button value=true groupValue=model.isDefault name="domain-type"}}
-                      {{#bs-tooltip placement="bottom" title='A default endpoint has an endpoint address in the format app-${id}.on-aptible.com.' bs-container=false}}
-                          Use app-{{ model.app.id }}.on-aptible.com default endpoint address
-                      {{/bs-tooltip}}
+                    {{#radio-button value=TYPE_ACME groupValue=vhostType name="domain-type"}}
+                        Use a custom domain with managed SSL.
                     {{/radio-button}}
                   </div>
                     <div>
-                      {{#radio-button value=false groupValue=model.isDefault name="domain-type"}}
-                        {{#bs-tooltip placement="bottom" title='Custom endpoints let you use your own certificate and domain name.' bs-container=false}}
-                            Use a custom endpoint address
-                        {{/bs-tooltip}}
+                      {{#radio-button value=TYPE_GENERIC groupValue=vhostType name="domain-type"}}
+                        Use a custom domain and a custom SSL certificate.
                       {{/radio-button}}
                     </div>
                 </div>
               </div>
-          {{/unless}}
 
             <div class="form-group">
-              <label class="block" for="vhost-type">Type</label>
+              <label class="block" for="vhost-type">
+                Endpoint Placement
+                {{more-info-icon
+                  title="Choosing an Endpoint Placement"
+                  bs-html=true
+                  body="
+                    <p>
+                    <strong>External endpoints</strong> are accessible via the
+                    a public internet. Use them for user-facing applications.
+                    </p>
+                    <p>
+                    <strong>Internal endpoints</strong> are only accessible by
+                    other apps in your private Aptible infrastructure. Use them
+                    for internal APIs.
+                    </p>"}}
+              </label>
+              {{#if placementNeeded}}
               <div class="radio">
                 <div>
                   {{#radio-button value=false groupValue=model.internal name="vhost-type"}}
-                    {{#bs-tooltip placement="bottom" title='External endpoints are accessible from the outside internet.' bs-container=false}}
-                        External
-                    {{/bs-tooltip}}
+                    External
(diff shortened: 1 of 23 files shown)


app/templates/stack/-certificate.hbs
```diff
@@ -2,12 +2,13 @@
...
+      {{/if}}
       {{#if certificate.inUse}}
-        {{#bs-tooltip placement="bottom" title="This certificate is currently in use and cannot be deleted." bs-container=false}}
-          <div>
-            <button class="btn btn-default btn-xs" disabled>Delete</button>
-          </div>
-        {{/bs-tooltip}}
+        {{! REVIEW: This used to have a tooltip, but it doesn't work on a
+            disabled object, and breaks layout. Are we OK losing it?}}
+        <button class="btn btn-default btn-xs" disabled>Delete</button>
```
sandersonet: As a workaround, could we add an `<i class="fa fa-question-circle"></i>` next to the button and apply the tooltip there? I could see this being confusing for some users.  We could also do that in a separate PR as I know there are several cases of disabled buttons with broken tooltips.

sandersonet:
👏 Fantastic work @krallin !

app/components/create-vhost/template.hbs
```diff
@@ -20,59 +20,122 @@
...
-          {{#unless serviceHasDefaultVhost}}
               <div class="form-group">
-                <label class="block" for="domain-type">Endpoint type:</label>
+                <label class="block" for="domain-type">
+                  Endpoint type:
+                    {{more-info-icon
+                    title="Choosing an Endpoint Type"
+                    bs-html=true
+                    body="
+                    <p>
```
fancyremarker: Awesome. Perfect explanation.

app/components/create-vhost/template.hbs
```diff
@@ -20,59 +20,122 @@
...
+
+                  <div>
+                    {{#if acmeVhostAllowed}}
+                      {{#radio-button value=TYPE_ACME groupValue=vhostType name="domain-type"}}
+                        Use a custom domain with managed SSL.
+                      {{/radio-button}}
+                    {{else}}
+                      <label class="disabled">
+                        <input type="radio" class="ember-view" disabled>
+                        Managed SSL unavailable: this feature is not available
```
fancyremarker: 👍  on "Aptible legacy infrastructure." Do we want to use that language elsewhere in place of "v1"?

fancyremarker:
This is really, really great work. Thank you!

app/components/create-vhost/template.hbs
krallin: Sure; I think I'm using "legacy stacks" in metrics. I can change that. Do we refer to v1 explicitly anywhere else? (I don't recall if we have a tooltip explaining why scaling isn't available on v1)

fancyremarker: Looks like only on [the support site](https://support.aptible.com/topics/paas/how-to-use-log-drains/#sts=Legacy Aptible “v1” stacks), actually — and only in one place. I'll change that now. I'm not really a stickler for consistency between "Legacy stacks" and "Aptible legacy infrastructure." If forced to choose, I prefer the latter. Up to you if you want to change language on metrics.

krallin:
@sandersonet pushed a commit to address your request to indicate why a cert cannot be deleted; looks like this:

image

@fancyremarker updated the copy in the metrics error message to:

```
No data available. If you just launched your containers, please try again later. Otherwise, please contact support (metrics are partially unavailable on Aptible legacy infrastructure)
```

app/components/create-vhost/template.hbs
```diff
@@ -20,59 +20,122 @@
...
+                This Endpoint type does not support internal endpoints. The
+                endpoint will be external.
+              </p>
+            {{/if}}
             </div>
 
-          {{#unless model.isDefault}}
+          {{#if acmeDomainNeeded}}
+            {{domain-input domain=model.acmeDomain valid=acmeDomainValid
+              tip="You'll neeed to be able to create a CNAME from this domain
```
krallin: @fancyremarker I'm not a huge fan of "endpoint address" — at least to me an address is something like `http://...`.

You're right that the repetition is awkward though. How's: 

> You'll need to be able to create a CNAME from this domain name to one provided by Aptible.

krallin: (@fancyremarker I pushed this for now; but let me know if you'd prefer something else)

fancyremarker:
@krallin: Can't comment on your copy change comment (commit was deleted), but LGTM.

chasballew:
@krallin - Suggested edits here: https://github.com/chasballew/dashboard.aptible.com/commit/52cb218268aa41ef69747b9e1959669924d92b6e

krallin:
@chasballew regarding your suggested edits:

I think this feature should have a name. For now, I had chosen this to be "Managed SSL", although I can understand why this name might not be ideal (more on that below).

That being said, I can understand your intention with some of the tweaks you've made, but the result is that we end up calling the same feature 3 different names in 3 different places:

-  ["Aptible-Managed SSL/TLS" there](https://github.com/chasballew/dashboard.aptible.com/commit/52cb218268aa41ef69747b9e1959669924d92b6e#diff-b187f452364c45a8be5a69fdd4ddd5faR65)
- ["Managed SSL/TLS" there](https://github.com/chasballew/dashboard.aptible.com/commit/52cb218268aa41ef69747b9e1959669924d92b6e#diff-b187f452364c45a8be5a69fdd4ddd5faR70)
- "Managed SSL" in the VHOST list (you might have missed that one though).

In my experience, features that don't have a name are difficult to talk about, which makes them a) difficult to document and b) difficult to market or even recommend. So, I'd like to retain a name for this feature. Do let me know if you disagree!

Assuming we agree that this feature should have a name, the question of what that name should be remains. It looks like you'd want us to steer away from choosing between "SSL" or "TLS" to refer to it (by using both). If so, then a good option might be "Managed HTTPS" — after all, this feature is currently only available for HTTP endpoints, and we do a little more than just provisioning a SSL cert (we automatically redirect HTTP to HTTPS).

Let me know what you think!

(other edits are fine; I'll incorporate all once we've resolved this)

Cheers,

krallin:
I added support for transitional certs for ACME, to support customers that are moving over an existing app to Aptible and want to reuse their cert while we set up ACME in the background. 

@fancyremarker @chasballew, you might want to review the copy. It's a little verbose, but I'm not sure we can explain the implications with less words.

Before releasing this, we still need to decide on a name... I think "Managed HTTPS" might be more appropriate than "Managed SSL". Let me know your thoughts!

krallin:
Screenshots:

image

image

krallin:
Note: after releasing this, I'll start working on "upgrading" an existing VHOST from Regular SSL to Managed SSL (and presumably the other way around as well).

chasballew:
@krallin - Let's ship now and edit the copy if it proves too confusing.

fancyremarker:
The copy for Transitional Certificates makes sense to me; I don't think any edits are necessary but we can make them later if anyone is confused.

PR #645, opened 2016-07-08
Fix CSP violations on sandboxes / staging

krallin:
We're hardcoding a lit of allwoed connect-src in our CSP, but it's
incomplete and doesn't work for sandboxes. This updates our CSP
connect-src rules to be dynamically constructed using the endpoint URLs
we're already dynamically generating.

(It also fixes a typo in `metricsBaseuri`, which should be
`metricsBaseUri` for consistency with everything else)

---

I didn't *remove* anything for the sake of safety.. but perhaps we should. Let me know. I reckon we don't enable this in prod (since e.g. api.aptible.com isn't in the list), so the impact should be pretty low. LMK if I'm missing something though.

Cheers,

### DIFF
--- a/app/services/metrics-api.js
+++ b/app/services/metrics-api.js
@@ -24,7 +24,7 @@ export default Ember.Service.extend({
     let accessToken = this.get("session.token.accessToken");
     let containerIds = containers.map((container) => container.get("id"));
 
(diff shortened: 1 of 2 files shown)


sandersonet:
👍

fancyremarker:
👍

Issue #642, opened 2016-07-07
Can orphan a user if they are removed from all existing role memberships


gib:
**Problem:** An organization user, "Joe", exists in only one role called "Developers". What happens to the "Joe" user when it is removed from the "Developers" role (when the role membership is destroyed)? 

Should the user also be destroyed after a confirm dialog is accepted?

gib:
addressed by https://github.com/aptible/dashboard.aptible.com/pull/679 which keeps users from removing a user's last role in an org

Issue #641, opened 2016-07-07
Dashboard users need to be able to remove a role


gib:
This process must ensure that the user is notified if this will orphan users when the role is deleted. If any of the deleted role's members are not members of any other role, they're orphaned.

Currently the select that provides users that can be added to a role is built from `organization.get('users')`, which would allow users with no other memberships to be added to a different role, but we may be assuming all users have at least one role. Should we be deleting users in this case? Asking the user that initiated the delete if they would like to delete those users? Hell of a confirm dialog... will need design / discussion.

PR #646, opened 2016-07-11
Add no organization error page

krallin:
When a user attempts to log in without being a member of an
organization, we currently throw a `TypeError`, which is problematic for
a few reasons:

- First, it doesn't let the user log out (without manually deleting
  cookies), which is bad from a security standpoint (they might not even
  realize they're still logged in).
- Second, it's very confusing: it's very unclear why they're receiving a
  `TypeError`. In practice this shouldn't happen very often, but it
  *can* happen if a user attempts to verify their email before accepting
  an invitation (aptible/auth.aptible.com#225 is
  evidence that some people do), in which case we should send them back
  on the happy path, i.e. tell them to accept the invitation.
- Third, it might start happening more often if / when we let users
  remove other users from their organization
  (#642,
  #641)

Obligatory screenshot: 

image

It's not *super* sexy, but I think it works. I'm a little bit concerned about that grey on grey being a little toned down (but that's arguably a problem for error pages?). Let me know what you think.

--

also, cc @wcpines, this should make the "accept-invitation" flow more foolproof.

### DIFF
--- a/app/error/controller.js
+++ b/app/error/controller.js
@@ -2,14 +2,14 @@ import Ember from "ember";
 
 export default Ember.Controller.extend({
 
-  title: function(){
+  errorCode: function(){
     return this.get('model.status') || this.get('model.name');
   }.property('model.status', 'model.name'),
 
-  subTitle: Ember.computed.alias('model.statusText'),
+  title: Ember.computed.alias('model.statusText'),
 
   message: function(){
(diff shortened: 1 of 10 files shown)


gib:
Nice!

wcpines:
Big improvement!

sandersonet:
👍

PR #647, opened 2016-07-13
Update ember-cli-aptible-shared to 0.0.93

krallin:
Oops; this was missing in https://github.com/aptible/dashboard.aptible.com/pull/635 — sorry about that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "aws-sdk": "^2.1.34",
     "broccoli-sass": "^0.6.6",
     "core-object": "^1.1.0",
-    "ember-cli-aptible-shared": "krallin/ember-cli-aptible-shared#revoke-tokens",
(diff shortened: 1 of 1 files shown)


krallin:
I dun goofed a little bit here. I released ember-cli-aptible-shared first for https://github.com/aptible/dashboard.aptible.com/pull/646 (as 0.0.92), but then merged https://github.com/aptible/dashboard.aptible.com/pull/635 first. Then, I released ember-cli-aptible-shared for #635, as 0.0.93

But, this means I can't add 0.93 here, since it needs #646 for the tests to work! And conversely, #646 won't build unless 0.0.93 is here, because the code on master needs it!

Anyway; that's not a big deal, it just means I needed to add 0.0.93 in #646, which is a little messy, but we'll survive.

PR #649, opened 2016-07-15
Fix broken test when creating a compliance user role

gib:
Updates the role type options component and fixes the test (thanks @sandersonet). The component now has a single action to update the role. A side benefit is the actions are tab key focus-able with button elements instead of divs.

Users are also directed to role members after creating instead of role environments which are not available for compliance roles.


sandersonet:
👍

PR #652, opened 2016-07-21
Only show service scaler success on success

krallin:
The service scaler shows a success message in a `finally` block, which
means it may show a success and an error at the same time if an
operation fails, whereas it should only show an error.

cc @sandersonet @gib @blakepettersson


sandersonet:
great, thanks!

PR #653, opened 2016-07-21
Upgrade ember-cli-sentry

sandersonet:
This upgrades our version of ember-cli-sentry to the latest version.  The key differences here are:

1) Raven.js is loaded via bower rather than CDN.
2) We no longer need a custom sentry service to catch exceptions

Also, the latest version isn't compatible with our older version of Ember, so I did have to overload the initializer and patch the service lookup. 

---

cc @krallin


krallin:
thanks @sandersonet !

I noticed a bunch of exceptions showing up in Sentry from aptible1.local; assuming that might be you? Is this something you've already looked into? (I suppose we don't want to report local exceptions into Sentry :) )

PR #656, opened 2016-07-25
Set referrer policy to origin-only

krallin:
This ensures we don't leak referrers to e.g. analytics third parties,
which is particularly relevant for e.g. pages whose URL includes an
opaque token, such as password recovery pages.


fancyremarker:
👍  Thanks!

sandersonet:
👍

Issue #669, opened 2016-07-29
Aborting signup flow before creating a BillingDetail leaves account in a bad state


sandersonet:
If a user aborts signup flow before committing payment info, their account is left in a bad state and their Dashboard is not operational (persistent 404 on BillingDetail).  Organizations missing a BillingDetail should be redirected to complete payment info.

sandersonet:
#670 was pushed to at least catch the 404 on the BillingDetail.  This prevents the redirect to the error screen and lets the dashboard work as expected for users that happen to have another valid organization.

gib:
- [ ] when a user only has one org with no billing detail, transition user to final payment step of welcome 
- [ ] when a user has multiple orgs and at least one has a billing detail, add a component to the side bar that gets the user to the payment step for that organization
- [ ] Red / Green refactor - This is a regression with the new role / permission checks that involved a check on the billing detail's plan

@sandersonet -- anything else to add here?

PR #670, opened 2016-07-29
Catch failed billing detail load

sandersonet:


sandersonet:
This PR addresses the most urgent aspect of #669 by preventing the entire Dashboard from just redirecting to an error screen when the billing detail fails to load.

PR #672, opened 2016-07-29
Billing detail redirects

### LINKED ISSUE #669: Aborting signup flow before creating a BillingDetail leaves account in a bad state

sandersonet:
This PR expands on #670 to fully fix #669.

* Any user who belongs only to one organization that lacks a billing detail will now automatically be redirected to complete payment setup.

* Any user who belongs to multiple organizations will have a new "Complete payment info" button next to each organization lacking a billing detail.

These somewhat minor fixes require a complete refactor of the payment-info screen as it previously just punted on multi-org setups by always using the first org found.   

This PR also renames the `requires-read-access` route to something more generic.  This is because that route is now used to filter and redirect on 3 possible bad/error states.

### DIFF
--- a/app/catch-redirects/route.js
+++ b/app/catch-redirects/route.js
@@ -2,6 +2,29 @@ import Ember from 'ember';
 
 export default Ember.Route.extend({
   redirect() {
+    return this.handleNoOrganizations() ||
+           this.handleMissingBillingDetail() ||
+           this.handleComplianceOnlyUsers() ||
+           true;
+  },
+
+  handleNoOrganizations() {
+    let organizations = this.modelFor('dashboard').organizations;
+
+    if(organizations.get('length') === 0) {
+      return this.transitionTo('no-organization');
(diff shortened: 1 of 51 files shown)


app/catch-redirects/route.js
```diff
@@ -0,0 +1,50 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+  redirect() {
+    return this.handleNoOrganizations() ||
+           this.handleMissingBillingDetail() ||
+           this.handleComplianceOnlyUsers() ||
+           true;
```
gib: nice readable methods

gib:
Thanks @sandersonet. This is awesome!

PR #673, opened 2016-08-01
Do not show invalid route in elevate

krallin:
This was reported a couple weeks ago via our bug bounty (and again over
this week end): the redirectTo parameter on the elevate route (which is
used to allow reloading this page) can be used to spoof the error
message that's shown.

While this is arguably reasonably unlikely to convincingly trick
someone, it's still worth fixing, which this patch does.

--

Fixes: https://trello.com/c/hNdb8op1/29-prevent-social-engineering-via-impersonate-s-redirectto
cc @fancyremarker @chasballew


sandersonet:
Nice, thanks for this

PR #674, opened 2016-08-01
Fix typo in connect-src

krallin:
We need a space between intercom and Metrics; looks like I messed this up in b33f6dab. It did use to work (perhaps because Chrome used to be more lenient about those errors?). Anyway, this should be the fix.


sandersonet:
:+1:

PR #676, opened 2016-08-01
Prevent potential issue with account creation

sandersonet:
This PR fixes two more bugs that came up today:

1) It's possible to inject blacklisted handle chars using our `handle-input` component. This is tricky, but, if you only ever input blacklisted characters, the sanitized value of the input is never persisted as the actual value.  This is because all of the sanitized characters are replaced with empty string and empty string == empty string. It's not immediately clear to me how to fix this, so I am temporarily replacing all blacklist chars with `-`.  This isn't ideal, but at least fixes the major issue of blacklisted characters hitting API and failing account creation. 

2) Users with 0 stacks (as in those affected by the above bug) are now redirected to a `no-stacks` error page.  This is a bad state that our UI can't gracefully handle without additional refactor.

### DIFF
--- a/app/components/handle-input/component.js
+++ b/app/components/handle-input/component.js
@@ -25,6 +25,6 @@ export default OneWayInput.extend(Autofocusable, {
       // FIXME: nonAlphaNumerics really should be replaced with an empty string,
       // however that breaks the `value` attr binding so that the first invalid
       // character is never actually stripped.
-      replace(nonAlphaNumerics, '-');
+      replace(nonAlphaNumerics, '');
   }
 });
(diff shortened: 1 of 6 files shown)


app/components/handle-input/component.js
```diff
@@ -22,6 +22,9 @@ export default OneWayInput.extend(Autofocusable, {
     return truncate(input, maxChars).
       replace(capitalLetters, replaceWithLower).
       replace(spaces, '-').
+      // FIXME: nonAlphaNumerics really should be replaced with an empty string,
+      // however that breaks the `value` attr binding so that the first invalid
+      // character is never actually stripped.
       replace(nonAlphaNumerics, '');
```
gib: Looks like a good solution.
Would doing the jiggery pokery (char replacements) on a copy of the value, then assigning it work?

PR #678, opened 2016-08-02
Refactor service slider component

sandersonet:
This purpose of this refactor was to fix the flakey app service scale acceptance test.  The cause of the intermittent failure is due to the race condition present in `showActionButtons` computed.  That computed attempted to verify that the slider has changed even though the `initializeContainerSize`computed actually updates the slider value to match--thus making `showActionButtons` extremely unreliable.  This PR removes the `initializeContainerSize` computed property and replaces many of the manual computeds with built-in macros.

This also removes all the `visibility` tricks and instead just doesn't render the buttons if they don't need to be there. 

--
cc @krallin

### DIFF
--- a/app/components/service-scaler/component.js
+++ b/app/components/service-scaler/component.js
@@ -8,50 +8,47 @@ export default Ember.Component.extend({
 
   isSliding: false,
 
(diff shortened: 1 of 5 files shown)


krallin:
Thanks for this @sandersonet ! : )

PR #685, opened 2016-08-07
Switch from IAM to ACM certificates

fancyremarker:
This does for our CloudFront distributions what aptible/pancake#124 did for our ELB certificate configurations, moving from IAM to ACM so that we can deprecate all IAM certificates (except for our one production EV certificate). The first two commits are just cleanup; see 99b0fd7 for relevant changes.


krallin:
I used this to set up the Dashboard for the brower testing Dashboard. Worked great.

fancyremarker:
(Replaced by aptible/dashboard-private#187.)

PR #688, opened 2016-08-10
Compliance dashboard

sandersonet:
This PR ports our Compliance Dashboard to Ember. To do this, I refactored all of our existing Sheriff data stores into a single service.  Going forward, this service should be used as the model for all compliance routes.


app/services/compliance-status.js
```diff
@@ -0,0 +1,102 @@
+import Ember from 'ember';
```
sandersonet: @gib - This is the service I mentioned earlier.  It encapsulates all of the requests needed to render almost all of the compliance screens.  Rather than calling the `store` in each route, you can instead just inject this service into a route and return it in the `model` hook.  All of the compliance status checks are isolated  into the `CriterionStatus` classes that this service initiates.

gib: Awesome -- thanks!

app/compliance-engines/index/route.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+  complianceStatus: Ember.inject.service(),
+  model() {
+    return this.get('complianceStatus');
+  }
```
gib: Skinny route. I'm with you on this. 💯

PR #693, opened 2016-08-12
fix log drain tests

sandersonet:
https://github.com/aptible/ember-cli-aptible-shared/pull/152 exposed some flakey tests in the log drain module.  These tests are particularly flakey because they are "guessing" at the number of assertions for a given test. It's not always guaranteed how many polling attempts will be hit for a given test.

This PR fixes that by limiting the polling attempt to 1, so the number of assertions can be guaranteed each test.


sandersonet:
Closing for now -- master doesn't yet reference a released version of shared(0.0.106+) that includes that fix.

PR #698, opened 2016-08-17
Fix UI regression

blakepettersson:
This fixes a regression found after the merge of #589.

cc @gib


gib:
Thanks @blakepettersson!

PR #717, opened 2016-09-05
Reload apps the same way as domains and databases

### LINKED ISSUE #715: VHOST provision_failed status copy could be improved
image

The copy says "... is provision_failed...", which might not be intentional.

cc @blakepettersson

blakepettersson:
Reload deprovisioning apps the same way we do for domains and databases
to be consistent with how they do it. This also means that we do not
need to reload the operation when deprovisioning an app, which should
also fix #715.

cc @sandersonet @krallin


app/models/app.js
```diff
@@ -17,6 +17,7 @@ export default DS.Model.extend(ProvisionableMixin, {
   lastDeployOperation: DS.belongsTo('operation', {async:true}),
   currentImage: DS.belongsTo('image', {async:true}),
 
+  reloadWhileProvisioning: true,
```
sandersonet: I wonder if we should overload `shouldReload()` to only reload while in `deprovisioning` as well? Since an app is can be `pending` for a long time and we don't use `provisioning` for apps?

app/components/list-apps/component.js
```diff
@@ -0,0 +1,39 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  sortBy: ['handle:asc'],
+
+  arrayChanged: function() {
+    Ember.run.once(this, 'checkModelArray');
+  }.observes('model.[]'),
+
```
sandersonet: The current best practice would be to do this without the use of the function prototype extension:

```javascript
arrayChanged: Ember.observer('model.[]', function() {...})
```

app/models/app.js
blakepettersson: That is true - the only statuses in use for an app are `pending` and `provisioned`, as well as 'deprovisioning'. 'pending' doesn't actually reload the model - only 'provisioning' and 'deprovisioning' do. 

I'll override `shouldReload()` (and refactor it a bit) since it will be a bit clearer on what's actually going on.

blakepettersson: @sandersonet I refactored `shouldReload()` so that it looks for a property called `reloadOn`. That's just an array of statuses which are the statuses the model will reload on. This also removes the `SHOULD_RELOAD` boolean, since if we don't want to reload the model whatsoever, we just send in an empty array.

The `reloadOn` properties are currently string arrays, since when I tried to send `STATUSES` I got a bunch of undefined errors. I'm guessing this has something to do with not exporting the `STATUSES` on the models?

app/models/log-drain.js
```diff
@@ -14,24 +9,9 @@ export default DS.Model.extend(ProvisionableMixin, {
   drainType: DS.attr('string'),
   drainUsername: DS.attr('string'),
   drainPassword: DS.attr('string'),
+  reloadOn: ['pending', 'provisioning', 'deprovisioning'],
```
sandersonet: Nice fix

sandersonet:
👍

app/models/app.js
blakepettersson: I fixed the issue with `STATUSES`; I added the enumeration in a separate file and imported it in each file that requires it.

PR #718, opened 2016-09-05
Some liquid fire improvements

sandersonet:
This PR updates the apps#create and databases#create route to actually be modal components.  

Sadly @blakepettersson this removes the current `list-databases` component and the need for the new `list-apps` component as you no longer need to bubble up the `forceRedirect` action.

Further, this PR fixes the issue of a user lacking `manage` permissions on an environment being redirected to the `new` route for apps/databases.  This is confusing as they can't actually create any new resources and they'd just get an error when attempting to create.  This PR hides the create button for these users.

This also changes the UX for empty index pages.  Rather than redirecting immediately to the `new` route, you now get a simple message with a button.  Images below.

#### Empty Index UI

screenshot 2016-09-05 16 40 38

#### Create DB UI
screenshot 2016-09-05 16 40 22


sandersonet:
I wrapped up this PR early in order to make it in time for Engineering Show & Tell, but ideally we'd use this same model on Log Drains and Certificates as well.

krallin:
> This PR hides the create button for these users.

Do you think we could show them a message explaining they don't have permission? In my experience just hiding the button when a user is missing permissions doesn't necessarily make it clear how they should go about resolving this problem.

sandersonet:
@krallin - Yeah that's a good point.  I'll change that behavior.

PR #719, opened 2016-09-06
Ignore expired_token on non-revoke logout

krallin:
When the user is attempting to logout, telling them their token expired
isn't particularly useful (after all; it's in the state they wanted it
to be...!). So, this patch ensures we don't throw an exception if the
user's token has expired.

However, we still throw one if the user was trying to revoke their
tokens. The assumption here is that silently failing might result in
them thinking their tokens were revoked when in fact they were not.

--

cc @sandersonet @gib


sandersonet:
👍

PR #724, opened 2016-09-09
Hide DB Connection URL behind reveal link

### LINKED ISSUE #634: Redact database URLs and warn when displayed
@sandersonet @gib @krallin : Customers semi-frequently copy/paste database URLs, including passwords, into email, help desk tickets, and other non-secure channels.

To help customers avoid this, we should:

1. Redact the full URL by default
2. Display a warning of some sort when the full URL is displayed or copied

One idea we discussed is a "Show URL" button or link that, when clicked, displays the URL and adds a red border and "SENSITIVE" badge to the UI.

sandersonet:
This PR hides our DB connection URLs behind a "Reveal" link

fixes #634

### DIFF
--- a/app/styles/_dashboard.scss
+++ b/app/styles/_dashboard.scss
@@ -187,7 +187,41 @@ header.resource-header {
       display: inline-block;
       margin-left: 6px;
     }
+
+    .reveal-link {
+      margin: 0;
+    }
+
+    &.db-connection-url-value {
+      min-width: 600px;
+    }
   }
+}
+
+.sensitive-indicator {
+  background: $color-danger;
+  font-size: 10px;
+  font-weight: $weight-semibold;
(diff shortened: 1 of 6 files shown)


app/styles/_dashboard.scss
```diff
@@ -187,9 +187,43 @@ header.resource-header {
...
       margin-left: 6px;
     }
+
+    .reveal-link {
+      margin: 0;
+    }
+
+    &.db-connection-url-value {
+      min-width: 600px;
+    }
```
sandersonet: I had to fix the db connection URL width to something really high.  Otherwise, in most cases, when the Reveal link is clicked, the element containing the connection URL will grow and wrap to the next line.  That is jarring, so instead I've hardcoded a large width so it should already be wrapped.

PR #734, opened 2016-09-14
Fix membership edit

sandersonet:


app/organization/members/edit/route.js
```diff
@@ -78,7 +88,7 @@ export default Ember.Route.extend({
           });
         } else {
           promise = role.get('memberships').then((memberships) => {
-            let userMembership = memberships.findBy('data.links.user', userLink);
+            let userMembership = memberships.findBy('user', user);
```
sandersonet: This is the key fix here.

PR #742, opened 2016-09-19
Only show visible database images

blakepettersson:
This fixes an issue where if a user has pre-existing database images
that are normally not visible, the select box should not have those
images available when creating a new database.

@krallin @fancyremarker @sandersonet


fancyremarker:
Thanks!! LGTM — @sandersonet, to you too?

PR #749, opened 2016-09-30
Show provision_failed status for default VHOSTs

krallin:
Default VHOSTs can fail to provision too (!), and we'd want to show the
error message in this case.

cc @blakepettersson @fancyremarker @sandersonet


sandersonet:
👍 Thanks for this

fancyremarker:
LGTM.

PR #754, opened 2016-10-17
Remove commented-out bits in crossdomain.xml

krallin:
We've received two invalid reports on security@aptible.com about this
file.

In both cases, the reporter probably ran some form of basic automated
scanning (or manual review) and didn't realize the vulnerable bits of
the policy are commented out.

While it doesn't make any functional difference, we might as well remove
out the commented out bits here to avoid receiving further invalid
reports.

--


fancyremarker:
Thanks!

PR #755, opened 2016-10-18
Reload app on invitation accept

krallin:
We cache organization permissions when loading the authorization
service, but when accepting an invitation, the user's permissions are
going to change.

We'd need to either invalidate the cached permissions, or invalidate
everything. This patch does the latter in the interest of un-breaking
accepting invitations, but the former might be a much cleaner approach.

--

cc @sandersonet @fancyremarker - as explained in Slack, I'm going to roll this out to un-break accepting invitations. I'm totally fine with reverting this or replacing it with a better fix.

### DIFF
--- a/app/claim/route.js
+++ b/app/claim/route.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import Location from "diesel/utils/location";
 
 export default Ember.Route.extend({
   model(params){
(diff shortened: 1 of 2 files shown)


krallin:
Build failed on unrelated tests that have flaked out before - merging.

PR #760, opened 2016-10-20
Update copy with product/role names; update role descriptions

chasballew:
@sandersonet - as mentioned, this is not tested (Watchman is unavailable and `ember test` is just hanging for me... 🤔 )


sandersonet:
Thanks :+1:

PR #763, opened 2016-10-25
Certificates: Create -> Upload

krallin:
The copy for uploading a certificate suggests you can create a new
certificate via Aptible, which could be misunderstood as signing a CSR
and returning a Certificate.

While the copy is technically accurate (we create a certificate record
in API), it's a little confusing, since you need to already have created
the certificate somewhere else to be able to use in on Aptible.

--

### DIFF
--- a/app/certificates/route.js
+++ b/app/certificates/route.js
@@ -17,7 +17,7 @@ export default Ember.Route.extend({
   },
 
   actions: {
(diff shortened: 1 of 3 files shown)


sandersonet:
Great fix, thanks @krallin

henrystewart:
Rad!

PR #764, opened 2016-10-26
Show certificate fingerprint

krallin:
If you have multiple certificates with the same hostname and issuer, it
becomes a little difficult to tell which one is which, especially if
they have issue dates that are close.

Fortunately, we do in fact extract the certificate fingerprint in API,
which can be used to easily tell two certificates apart. This change
adds a truncated fingerprint (7 characters, which is what git uses for
truncated fingerprints too - it's not like we're relying on this for
security, just convenience), which is shown on the certificates list
page and the selector.

I also updated the list page to explain why some certs have no issuer
(they're self-signed), and to fallback to issuerCommonName if
issuerOrganization is missing (the former is guaranteed to be present if
the cert was signed by a CA, the latter is not).

I did have to make the certificate name a little more succinct (remove
some labels), or Chrome wouldn't always show the fingerprint if the
option label was too large in the cert picker.
## 

Some screenshots:

image

image

### DIFF
--- a/app/models/certificate.js
+++ b/app/models/certificate.js
@@ -2,6 +2,8 @@ import DS from 'ember-data';
 import Ember from 'ember';
 import { formatUtcTimestamp } from '../helpers/format-utc-timestamp';
 
+const FINGERPRINT_DISPLAY_SIZE = 7;
+
 export default DS.Model.extend({
   // properties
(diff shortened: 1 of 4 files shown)


fancyremarker:
👍

PR #765, opened 2016-10-26
Release ALBs!

krallin:
This shows the platform for a given Endpoint. We shouldn't release this
until ALB is actually the new default in API.
## 

Screenshot (happy to tweak the copy):

image


fancyremarker:
👍 , explanation looks good.

krallin:
I'm upgrading this PR a little bit. Rather than just show the platform, customers can now pick their platform... and upgrade to ALB! This uses the new fancy modals with a little help from @sandersonet 😄 

Screenshots:

Endpoint List:

image

Upgrade modal:

image

New Endpoint menu (this is starting to be a little crowded 😢 ):

image

app/app/vhosts/new/route.js
```diff
@@ -11,7 +11,7 @@ export default Ember.Route.extend({
     var stack = app.get('stack');
 
     return Ember.RSVP.hash({
-      vhost: this.store.createRecord('vhost', { app }),
+      vhost: this.store.createRecord('vhost', { app, platform: 'alb' }),
```
sandersonet: If they all default to `alb` now, we could just set this as the `defaultValue` for the `platform` attr directly on the model.

krallin: Nice; thanks! Updated.

PR #768, opened 2016-11-03
Only allow ELBs on v2 😢

krallin:
Unfortunately, legacy stacks can't support ALBs unless we add a second
public subnet (without any instances in it). Considering these stacks
are in maintenance mode at this point, it's preferable to just disallow
this option rather than make last minute changes (also, in practice,
ELBs aren't as bad there as they are on v2).

---

(It's regrettable I didn't catch this before, but our only v1 test stack is a bastardized v2 turned v1 stack... which had multiple subnets 👎)

### DIFF
--- a/app/app/vhosts/new/route.js
+++ b/app/app/vhosts/new/route.js
@@ -7,8 +7,8 @@ export default Ember.Route.extend({
   },
 
   model() {
-    var app = this.modelFor('app');
-    var stack = app.get('stack');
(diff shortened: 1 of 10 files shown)


fancyremarker:
I agree this is the right fix, thanks. I'm not sure without further investigation how the alternative (adding additional public subnets/AZs to v1 stacks) would affect/break functionality and v1 assumptions.

PR #769, opened 2016-11-03
Allow updating and deleting a failed Endpoint

krallin:
We need to allow updating failed Endpoints, because with ALBs customers
must provide a certificate that's within its validity period, or it'll
be rejected when we try to upload it into IAM.

It's probably also a good idea to let customers delete these Endpoints
if they realize they've made a mistake (right now, you can't delete the
Endpoint unless you first get it to succeed provisioning, which can be a
little hard since we now require that the health check passes when
provisioning the Endpoint).

---

### DIFF
--- a/app/app/vhosts/edit/route.js
+++ b/app/app/vhosts/edit/route.js
@@ -41,6 +41,13 @@ export default Ember.Route.extend({
       let stack = this.currentModel.stack;
       let certificatePromise;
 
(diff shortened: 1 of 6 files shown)


fancyremarker:
Big improvement, also for this reason you mentioned:

> It's probably also a good idea to let customers delete these Endpoints if they realize they've made a mistake

LGTM, I figure @sandersonet will agree but leave it to him to comment on Ember details.

PR #772, opened 2016-11-07
Use assert.hasElement for app and db sidbar tests

bantic:
In the course of fixing some JSHint errors in the tests I noticed a small JavaScript gotcha that was causing some tests to redundantly pass. The `find` test helper function returns `[]` when it doesn't find anything, but `[]` is truthy, so assertions like: `assert.ok(find('.not-present'))` were present and passing.

This PR adds a `assert.hasElement(selector, message)` assertion that I will introduce to other tests as I come across them.

cc @sandersonet


bantic:
I `skip`'d two tests that were false-positive passing before but actually are failing now. I did a bit of investigation to determine why and will dig into it more tomorrow.

sandersonet:
👍 Thanks for this @bantic!

PR #776, opened 2016-11-09
Upgrade Ember-CLI

mixonic:
This puts the codebase on a path toward using modern addon APIs, and eventually toward updating Ember-Data, Ember and getting engines.

Todo:

* [ ] Still a bunch of linting issues largely caused by the jshint change. Seems easy to add the "globals" to each file instead
* [ ] Error loading gridiron `Error while processing route: gridiron-admin.index Assertion Failed: Cannot delegate set('organizations', ) to the 'content' property of object proxy (generated gridiron-admin controller): its 'content' is undefined. Error: Assertion Failed: Cannot delegate set('organizations', ) to the 'content' property of object proxy (generated gridiron-admin controller): its 'content' is undefined.`
* [ ] Confirm CI works


public/crossdomain.xml
```diff
@@ -2,5 +2,14 @@
...
+
+  <!-- Most restrictive policy: -->
   <site-control permitted-cross-domain-policies="none"/>
+
+  <!-- Least restrictive policy: -->
+  <!--
+  <site-control permitted-cross-domain-policies="all"/>
+  <allow-access-from domain="*" to-ports="*" secure="false"/>
+  <allow-http-request-headers-from domain="*" headers="*" secure="false"/>
+  -->
```
krallin: Can we not add these back? Rationale in: https://github.com/aptible/dashboard.aptible.com/pull/754

mixonic: Ha, weird. Yup I'll drop 'em.

krallin: Thanks! 😄

tests/acceptance/apps/sidebar-test.js
```diff
@@ -56,7 +56,7 @@ test('includes breadcrumb to parent stack', function(assert) {
   signInAndVisit(`/apps/${appId}`);
 
   andThen(function() {
-    let stackHandle = find(`h3:contains(${stackHandle})`);
+    let stackHandle = find(`h3`); // TODO this should read :contains(somehandle)
```
bantic: This should be ok after a merge. It was addressed in #772. In addition to the JSHint error the assertion below is a false-positive (it is essentially: `assert.ok([], 'has stack link')`).

